PR #254, opened 2018-01-30
Fix BaseURL when hosted behind a TLS reverse proxy

### LINKED ISSUE #253: Wrong base URL when hosted behind a TLS reverse proxy
https://github.com/getfider/fider/blob/master/app/pkg/web/context.go#L269
```golang
//BaseURL returns base URL as string
func (ctx *Context) BaseURL() string {
	protocol := "http"
	if ctx.Request.TLS != nil {
		protocol = "https"
	}
	return protocol + "://" + ctx.Request.Host
}
```

When hosted behind a TLS reverse proxy (without TLS between proxy and Fider), `Request.TLS` is `nil` and therefore it leads to:
- Mixed HTTP / HTTPS
- A blank page for the main page (`history.replaceState` will not work with a wrong URL).

A potential solution would be to set a new `BASE_URL` environment variable.

Hellslicer:
It fixes #253 by adding a check for the standard `X-Forwarded-Proto` header.

### DIFF
--- a/app/pkg/web/context.go
+++ b/app/pkg/web/context.go
@@ -266,7 +266,7 @@ func (ctx *Context) ActiveTransaction() *dbx.Trx {
 //BaseURL returns base URL as string
 func (ctx *Context) BaseURL() string {
 	protocol := "http"
-	if ctx.Request.TLS != nil {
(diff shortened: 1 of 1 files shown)


goenning:
Thanks for the contribution!

Hellslicer:
Thank you for the merge and more generally for Fider, cheers!

PR #316, opened 2018-03-12
Allow unauthenticated SMTP requests

LukeSheard:
Following up from this post, https://feedback.fider.io/ideas/74/allow-unauthenticated-smtp-for-self-hosted-instances, this has been working for me. 

*I'm not sure how to test this though, so any review comments on that front are welcome.*


goenning:
I don't think that'll be enough. I'm looking at Go source and I see that `smtp.SendMail` function accepts nil as second parameter, in such cases, it'll skip the authentication step. See https://golang.org/src/net/smtp/smtp.go?s=9230:9308#L309

My suggestion is send `nil` if both username and password are empty and avoid calling `smtp.PlainAuth`.

app/pkg/email/email.go
```diff
@@ -11,6 +12,14 @@ import (
 
 var cache = make(map[string]*template.Template, 0)
 
+//Auth creates an auth object for sending mail
```
goenning: If you can just move this function to the `smtp` package we can get this merged. The `email` package is supposed to be a common ground for all different sort of email providers and this method is very specific to SMTP.

LukeSheard: My golang skills couldn't work out how to do this... Any pointers for reading for understanding what's going on?

goenning: Every folder is called a package by the name after the last slash. So app/pkg/email is the email package. app/pkg/email/smtp is the smtp package. 

It’s organized in such a way that we can have one package for each email provider (smtp, mailgun, sendgrid, etc), while the email package contains shared structs/methods common for all providers.

So when I say to move things from email package to smtp package, is to basically move the “Auth” function to the app/pkg/email/smtp folder, inside the smtp.go file.

You might have noticed that the Go stdlib has a smtp package already, so that’s why in the beginning of the smtp.go file I imported that package and renamed to gosmtp, so you’ll need to call the stdlib function like  gosmtp.Auth(...)

Did it help? 😀

goenning:
@LukeSheard do you plan to continue on this? Otherwise I can jump in and finish it, I'd like to include this on the next version. Thanks!

LukeSheard:
@goenning Sorry I got completely sidetracked by work. I had a little go at making the changes you asked for but my Go knowledge hit a wall. Would you be okay with finishing it off? Apologies!

goenning:
@LukeSheard sorry to bother you again, but can you enable the "allow edits from maintainers"? this will enable me to continue this PR instead of creating a new one (and losing history). Thanks!

LukeSheard:
@goenning It's already on in my view!

PR #546, opened 2018-10-01
Add .nvmrc

ashishmohite:
-  improves build reproducibility by switching to an appropriate node version


goenning:
Thanks for your contribution. I like the idea of having a .nvmrc as some use it a lot, although I’m not a big fan of adding it as a dependency to the contributing. The less tool we require, the less friction it creates for new contributors to join.

ashishmohite:
@goenning can I add optional in parenthesis in front of the requirement? or should just remove it from the requirements ?

goenning:
I’d prefer to revert that file and add a note beside the Node.js requirement, something like:

- Node.js 10+ (https://nodejs.org/ or run `nvm use` if you have nvm)

What you think?

ashishmohite:
Done

goenning:
@ashishmohite merged. Thanks! 😀

Issue #425, opened 2018-06-10
Gracefully handle erros when rendering the UI


goenning:
**Problem:** when an error is thrown during the rendering on the UI, the whole tree crashes and a blank page is presented. We have logs on console, but the user doesn’t receive any feedback.

**Solution:** React 16 introduced a new API to handle errors. We should make use of it on the top level rending declaration and catch all errors.

If the user is using IE 10 or earlier version, we should add a message that the browser is not supported. If not, then we should show a message that something went wrong and we couldn’t display the page. There’s no need to redirect, we just need to prevent full white pages.

goenning:
Hey @cfilby, you're more than welcome to take it!

I agree with 1 and 2. Although for item 3, I believe we could open another issue and a separate PR to not mix up these two subjects. 

In regards to styling, we already have a 500 page for errors that happens on the server-side, so maybe we can reuse the same design for client-side errors. At this stage, anything better than a white page would be great 😀

PR #549, opened 2018-10-01
Add ErrorBoundary and ShowError to handle React exceptions.

cfilby:
**Issue:**  #425 Wraps React application with ErrorBoundary and creates new ShowError component to present a generic error page when an exception is caught.

This presents a pretty barebones error page modeled after the SignUp page. I've added some unit tests to try and verify the ErrorBoundary is properly catching errors from children. These are currently somewhat slow due to the use of mount. At present, the Error Message is purely generic since it seemed like it would be poor UX to display that level of exception to an end user. Is there anywhere you'd like me to log/display this error? 

Thanks! Please let me know if there's anything I can change on my end.


public/components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+  public componentDidCatch(error: Error, errorinfo: React.ErrorInfo) {
+    this.setState({
+      error,
+      errorinfo
+    });
+  }
+
+  public render() {
+    const { error, errorinfo } = this.state;
```
goenning: You can remove `errorinfo` from here if not using it.

goenning:
Amazing PR, thanks @cfilby

Regarding your question, we currently use Google Analytics to store events and exceptions. Search for `analytics` to find it in the code. At some point I'd like to remove Google Analytics for privacy reasons, but that's what we have for now, so please use `analytics.error` to send out the exceptions to GA.

I have one suggestion here. It would be awesome to show the exception on the ShowError component if the application is running on development mode. This would be a great DX feature. You can check that with `!Fider.isProduction()`. What do you think?

cfilby:
Thanks! I agree with your recommendation - I didn't know there was already a way to check the current environment. I'll update the code today/tonight to conditionally render the error message and conditionally report the error to Google Analytics if it's in production mode.

I'll remove errorinfo as well - I left that in there until I could confirm what error information you wanted reported.

goenning:
To be honest, I don't know what kind of information errorinfo holds. As we're now going to show the error on that component, we might include this as well if that's useful for the developer.

cfilby:
I've tried updating the PR in the following ways:

* Add two new props, showError and onError. When showError is true it will pass the error message to ShowError, otherwise it allows ShowError to render a generic message. onError is a callback prop that passes the error upwards for handling. 
* Pass !Fider.isProduction() and a logProductionError method to the ErrorBoundary in index.tsx
* Tried adding a few more tests to better cover this behavior. 

At the moment the errors I'm receiving are obfuscated due to it being considered a [Cross Origin Error](https://reactjs.org/docs/cross-origin-errors.html) in the development environment. I haven't looked into how exactly the app is being served up yet to assess which of those fixes would make sense.

Let me know if there's anything else I can change to improve this! If there's more expressive ways I can express this in Enzyme I'd be more than happy to refactor my tests.

goenning:
Alright. I liked the onError props! 

As with regards to showError, my idea was a bit different. 

1. I think we should always pass both Error + ErrorInfo objects to ShowError component. This component should be smart enough (by using the Fider.isProduction) to know when to expose the details of the error or not.

We can also keep both title and message as props like you had before. Inside ShowError component we can display the error with: 

```
<pre>
  {props.error.toString()}
  {props.info.componentStack}
  </pre>
```

This should format the error nicely!

2. I'm also thinking that if you use the css classes `container failure-page` like we do on 500.html, you would get same styles from server error pages out of the box.

3. Last one :) You can use `<TenantLogo size={100} />` instead of that hardcoded image, this will display whatever Logo that tenant has.

goenning:
I forgot one other minor thing. Does React still logs the error to `console.error` when you catch it with `componentDidCatch`? I think we should always log errors to console.error, on both PROD and DEV mode.

cfilby:
1) Sounds good. I was initially making ShowError as more of a presentational component and keep it ignorant of higher level details like the Fider library. I've changed ShowError to also accept an error, errorInfo and showError prop. The last prop is mostly included to simplify testing so the Fider.isProduction() check can be swapped out. The showError prop probably could use a better name though.

2) Done. 

3) Done. 

4) By default React will log all errors in Development mode to [console](https://reactjs.org/blog/2017/07/26/error-handling-in-react-16.html#component-stack-traces). I've added a console.error statement to the logProductionError method so the error is printed in production as well. 

Let me know if there's anything else you'd like me to edit! Thanks for your patience in providing feedback.

cfilby:
I guess in hindsight I could have easily stubbed the Fider.isProduction method on the global object for testing purposes. My personal preference is having an explicit prop for that decision logic to leave the component more configurable. That said, I'm fine with it either way you'd want it.

goenning:
You’re totally right. I was aware that unit testing would be more complicated with that global “Fider” object. You’re solution is fine! We might one day replace this global object with a ReactContext, that should make testing easier.

As for the name “showError”, I think there’s no better name for it, it’s named after what it does. Although I think what is confusing is having the name of the prop same as the component name. So maybe the ShowError component could be renamed to ErrorPage, as it’s actually a full page and not a reusable component. (That’d also change the folder and the css class name prefix to “p-“)

I actually thank you for all the patience you have going through my comments! I really appreciate the time you’ve put into this! And if you disagree with anything, please let me know. I’m always happy to discuss and get to a common agreement 😀

cfilby:
I went ahead and converted ShowError into the ErrorPage component and updated the styles accordingly. Let me know if anything else is missing! 

And no problem, happy to help!

goenning:
@cfilby It's merged! Thanks for your contribution!

Issue #522, opened 2018-09-04
Store events to allow further protection


goenning:
The idea is to create a SQL table to store all sorts of events that happen on Fider. This table would store: Client IP, Event Name and Date. Note that the User would not be stored so that we don't have a link between User and the IP.

Example of events would be: new post, new vote, delete vote, new comment, edit comment, page view, sign in, etc...

We can then mitigate spam / attacks based on Client IP and Event. How that'll be done is yet to be defined. This issue will be used for storing data only.

cfilby:
I'm interested in this one. I'm somewhat new to Go so I want to make sure I understand what I'd need to do to satisfy this PR. From what I've seen in the code, it looks like I'd need to:

* Add a Migration to create the Event Table. This table would contain an id, client_ip, name, created_at (string)
* Create an Event Model.
* Create an EventStorage interface.
* Add EventStorage implementations for in-memory and Postgres.

Is that correct? 

What sort of CRUD methods other than Add would you expect? GetAll? GetByClientIP? Should the event name be free text at the storage level, or would you want named events?

Thanks!

goenning:
@cfilby all your bullet points are correct, although created_at would be a timestamptz and not a string.

I'm thinking that we could use this data for analytics purposes too, like plotting a chart of how many votes/comments/posts have been created per day for the last X days. In that case, we'd need a tenant_id (with its FK), pretty much like all other tables.

This issue is just about collecting the data, so for now the EventStorage could have a Store (or Save) method. We can create the Read operations later on. 

I think that each consumer should be responsible for naming the event. I don't expect that we will have the same event name being generated in multiple places, so that'd not generate any duplicate code.

As for the names, I like what DigitalOcean does, so maybe we could copy it (see picture below)? Our events would be like: `post.create`, `post.update`,`post.respond`, `post.vote`, `post.unvote` (is there such word? 😀), `comment.create` and so on...

<img width="157" alt="screenshot 2018-10-02 at 18 58 48" src="https://user-images.githubusercontent.com/94755/46367368-41ab0380-c675-11e8-800d-f7a24dec3ddb.png">

cfilby:
Whoops! Meant to put that (string) next to the name, not the created_at column. 

I think that resource.action format makes plenty of sense. I'll go ahead and start working on this, thanks!

PR #551, opened 2018-10-03
Add Support for Persisting Audit Events

cfilby:
**Issue:** PR for #522.  Add basic event storage schema and functionality.


cfilby:
I believe this is ready for an initial review. This PR does the following:

* Creates a new Events migration with the following fields: id, tenant_id, name, created_at. 
* Creates an Events interface with two methods, Add and GetByID (named to be consistent with the other Add storage interface methods).
* Adds inmemory/postgres implementations for event storage with tests for the postgres implementation. 

The two questions I had were:

* Is it worth embedding the base interface? At the moment SetCurrentUser won't have any real effect. I ended up doing it since I figured it'd make things upstream easier if they were depending on the Base interface for common functionality.

* Would you recommend creating a separate dbModel for Event? Or given the simplicity of the model is it fine to reuse it at both layers? 

As an aside, is there a script somewhere to generate new empty migration files? I tried looking around a bit and just ended up generating the filename manually this time around.

Let me know if you see anything that I can change or improve!

goenning:
Hey!

1. The base interface is easy to implement and will fit better into current structure, so I’d recommend to implement it like you did.

2. No need for the dbModel. My only suggestion would be to remove the TenantID from the model. Saving/querying the tenant_id should be done inside the storage struct (like you did), but we don’t need to expose that as a field.

3. You forgot the ClientIP field 😀 You can get that from the web.Context struct. I’m now wondering if we should inject this as a dependency on the EventStorage constructor. If we do that, we can keep the Add method with a simple signature, just the name of the event as a parameter.

4. We don’t have a migration generator, I usually just create a new file manually

cfilby:
1. Cool, that's what I was figuring.

2. Fair enough, I went ahead and removed that from the Event model. 

3. Whoops! That's what I get for coding late at night - sorry about that. I've added the clientIP INET field to the existing migration. Right now this is just an argument to the new event function. 

I think whether or not to inject the context depends on the overall lifecycle of the storage objects. If you'd expect to create new storage objects with each request or if you wanted cancellation, then it'd probably make sense. I haven't used with Contexts much yet, but to me passing the Context directly to the storage does expose it to a lot of high level implementation details, whereas just passing the clientIP keeps the storage interface simple and ignorant of the web layer above it. I'm good with either approach though, just let me know. 

4. Ah okay, just wanted to make sure I wasn't missing anything!

As an aside, it looks like the WrapRequest method manually extracts the IP from the forwarding header. The net package has a similar function that might be worth [looking at](https://golang.org/pkg/net/#SplitHostPort) depending on your needs.

Thanks!

cfilby:
Sorry, did another commit to try rerunning the CircleCI tests. It looks like it was failing due to timeout issues or something.

goenning:
Are you talking about this line here?

```
strings.Split(request.RemoteAddr, ":")[0]
```

I never though about using SplitHostPort, but I guess it'd fit well. If you want to give it a try, just use a separate PR.

Now that we're talking about Client IP, I also noticed that if we can't parse the IP, we default it to "N/A". If this happens, we won't be able to insert into events table, because client_ip is `inet` type. This will most likely fail the request with a 500 response.

I think we have two options here:

1) We remove the "N/A" default value and allow null into client_ip field. The Add function would then have to check whether client_ip is empty or not;

2) Change client_ip to string;

Any thoughts?

cfilby:
Sure! I can go ahead and do that. Looking at the source code it seems like it may catch more edge cases (although it has the potential to return an error). Is the expectation that Fider will always run behind a reverse proxy? Right now it looks like the clientIP is pulled from the X-Forwarded-For header, so if you run it locally the client IP looks like it will always be null. While making that PR I could go ahead and have it fallback to the RemoteAddr property of the request if the X-Forwarded-For header is empty.

Do you mostly see the events table being used to prevent system abuse, or do you also envision it as a general purpose log? If you're trying to prevent abuse then you may want to leave the ClientIP mandatory and reject the request if the ClientIP can't be parsed. If we wanted to make ClientIP optional it seems like we'd want to keep INET field if possible for the stricter validation and drop the NOT NULL clause. My guess is that I could try to create a dbModel for events that uses a NullString as the datatype to make the IP optional seeing how we're currently just passing Postgres a string with the IP anyway.

I'm good with any of the approaches above though (Mandatory INET, Optional INET, Optional String), so let me know what you think and I'll go ahead and make those changes.

goenning:
We already parse from RemoteAddr, see: https://github.com/getfider/fider/blob/8410e5835c5124c931490aff05d8aa4f5429580d/app/pkg/web/request.go#L49-L57

It'll be a general purpose event log, which can then be used for multiple ends. So let's make it an Optional INET. The goal is to always save the Client IP, but if for any reason it's not available, so be it, save as NULL but don't fail the request.

I also think we can remove the default "N/A" when ClientIP is empty (line 53 above). That doesn't add any value, just make everything harder.

cfilby:
Haha, whoops, definitely overlooked that one (embarrassingly, I added some debugging statements before the block with RemoteAddr but didn't look in the block XD). 

The clientIP field should now be nullable and I've removed N/A as the default case.

One thing I noticed is that some of the other models are using NullableString for the retrieval, but not for the insert. I think you may want to use those for the insert as well if you want empty string values to be NULL in the database - my empty ClientIP string test failed on insert due to passing "" instead of NULL. 

Let me know if there's anything else I can tweak!

goenning:
When the database was created, I decided to use nullable varchar for certain fields in the database. But in 99% of the our use cases, a nil string and an empty string have the exact same meaning. I then realised that, because Go doesn't have nil strings, it's easier to treat the database fields just like any other Go string, without null values.

AFAIR, We use NullableString because some queries might not return all fields and `dbx` package doesn't convert null strings to empty strings.

goenning:
Thanks for the PR! 🎉

Issue #519, opened 2018-09-03
Some browsers are not persisting  __oauth_identifier cookie


goenning:
When this happens, the user is unable to sign in with OAuth. This cookie is used for security reasons based on OAuth requirements for parameter 'state'.

It seems like it's only happening to Google Chrome.

PR #552, opened 2018-10-05
tooling: update to Go 1.11.1 and enable SameSite=Lax

### LINKED ISSUE #509: update to Go 1.11

goenning:
**Issue:** This is an attempt to #519 by enabling `SameSite=Lax` (reference [chrome bug 696204](https://bugs.chromium.org/p/chromium/issues/detail?id=696204))

Even if it doesn't fix the bug, it's good for security reasons and it's also a Go version bump.

closes #509


goenning:
Looks like we're blocked with until 1.12 is released https://github.com/golang/go/issues/26186

Issue #554, opened 2018-10-06
Replace Semantic-UI-CSS Icons with React-Icons


goenning:
Two reasons for this change:

1. The goal is to completely remove Semantic-UI from Fider;
2. We only use a few icons from Semantic-UI-CSS, but bundle everything;

I'm not sure if [React-Icons](https://github.com/react-icons/react-icons) is the best choice, but it looks promising. We're open to other options as well. 

The goal is to replace every single Icon we use with the same icon from React-Icons. We also need to compare previous bundle vs new bundle.

For reference, this line needs to be removed from `index.tsx`

```
import "semantic-ui-css/components/icon.min.css";
```

ganesharulanantham:
@goenning Will take this up
List of icons used from semantic-UI-CSS in our project. Will check React Icons whether all suitable icons exist. Until, do you have any thoughts to share on this?

1. add
2. announcement
3. cancel
4. caret up
5. check
6. check circle
7. comments outline
8. delete
9. dropdown
10. edit
11. exclamation triangle
12. lightbulb outline
13. lock
14. play
15. remove
16. save
17. setting
18. times circle
19. user
20. volume off
21. volume up

ganesharulanantham:
Corresponding Font awesome icons available in React icons in same order .
@goenning Just compare and let me know whether I can go ahead and make the changes

1. fa-plus
2. fa-bullhorn
3. fa-times
4. fa-caret-up
5. fa-check
6. fa-check-circle-o
7. fa-comments-o
8. fa-times
9. fa-caret-down
10. fa-pencil-square-o
11. exclamation triangle (not found)
12. fa-lightbulb-o
13. fa-lock
14. fa-play
15. times
16. fa-floppy-o
17. fa-cog
18. times circle (not found)
19. fa-user
20. fa-volume-off
21. fa-volume-up

goenning:
Nothing to add. Go for it! Thanks for the help! 😄

For times circles, I found the `FaTimesCircle` on React Icons

goenning:
@ganesharulanantham are you still interested on this issue? It’s ok if you are and need more time! But if you don’t want it anymore, just let me know. I’d like to move this forward and willing to do it myself 😀

Thanks!

PR #556, opened 2018-10-07
Refactor ClientIP extraction logic

cfilby:
**Issue:**  The ClientIP logic didn't use the existing net.SplitHostPort function. 

This PR refactors the ClientIP extraction logic to better express the intent/flow of how the ClientIP is determined and uses the built in net.SplitHostPort. SplitHostPort supports a wider range of potential IP Address formats.

### DIFF
--- a/app/pkg/web/request.go
+++ b/app/pkg/web/request.go
@@ -2,6 +2,7 @@ package web
 
 import (
 	"io/ioutil"
+	"net"
 	"net/http"
 	"net/url"
 	"strings"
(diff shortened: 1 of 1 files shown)


cfilby:
Also, let me know if there's anything else you want me to look into/refactor elsewhere in the project - I'd be happy to help!

goenning:
Nice one, thanks!

I want to completely remove `semantic-ui-css` and `semantic-ui-react` from Fider. We only use a single component/resource from each package, so it's not worth the cost of extra bytes on the bundle. This has been a long time effort and I'm very close to the goal. I'm currently working on removing `semantic-ui-react`. Issue #554 should remove `semantic-ui-css`. If you want to have look, it'd be great. 

After that, the next front end challenge will be code-splitting. No idea how I'm going to do that, but it'll be fun. Happy to have you onboard! 😀

PR #561, opened 2018-10-09
Update .example.env

cameronmoreau:
**Issue:**
`EMAIL_NOREPLY`, `EMAIL_SMTP_HOST` and `EMAIL_SMTP_PORT` are required to start the app or else `mage watch` will panic.

Added `null` to these fields in the `.example.env` to help new contributors start the app according to the [contribution guide](https://github.com/getfider/fider/blob/master/CONTRIBUTING.md#getting-started-with-fider-codebase) steps.


goenning:
Good catch, thanks!

I’m not sure if we should set as null or put some dummy config, like smtp.yourdomain.com, user@yourdomain.com and yourpassword

But for now, null is better than empty 😀

Issue #544, opened 2018-09-30
markdown preview


goenning:
**Source:** https://feedback.fider.io/posts/29/preview-of-markdown-field

**Problem:** Fider accepts markdown text on the textarea for a new idea and comment, but the users can't see how the text will look like until they submit it. 

**Solution:** We could implement something like GitHub, where there is a tab to preview the text. Ideally this should be made as a React component that we can use in both places. We already use `markdown-it` NPM package, so this is a front-end only task.

ganesharulanantham:
@goenning will take it up

goenning:
I remember having this issue before. Aside from `mage watch`, do you have many other things open? You can also increase max files with this: https://superuser.com/questions/433746/is-there-a-fix-for-the-too-many-open-files-in-system-error-on-os-x-10-7-1

ganesharulanantham:
@goenning thx for the help.
Landed pull request [https://github.com/getfider/fider/pull/562](https://github.com/getfider/fider/pull/562)

RokeJulianLockhart:
> https://github.com/getfider/fider/issues/544#issuecomment-549092505

@Stale, this is still a problem. I just posted a comment at [`ideas.obsproject.com/posts/2737`](https://ideas.obsproject.com/posts/2737/provide-a-replacement-for-the-deprecated-system-theme#:~:text=%3E%20Just%20as%20a%20note%2C,those%20are%20being%20publicly%20tracked?) with CommonMark blockquote syntax, which wasn't rendered. There was no way to determine this before I posted it.

It's rather ironic that even https://github.com/getfider/fider/pull/562#event-2785702656 was closed due to being stale. You should reconsider your usage of that bot.

PR #562, opened 2018-10-10
markdown-preview-added

ganesharulanantham:
**Issue:**  markdown preview https://github.com/getfider/fider/issues/544

To preview the markdown text input given by user in new idea description, comment and welcome message. Created a new React component wrapping Text area and Markdown Preview. 
Example Screenshot
screenshot 2018-10-11 at 12 03 29 am
screenshot 2018-10-11 at 12 03 35 am


goenning:
I have a few comments, hope you don't mind!

I have just removed `semantic-ui-react`, this has been a very long process and I finally merged the last PR. This has dropped the css from 172 KiB to 114 KiB, and the js from 378 KiB to only 105 KiB! I'm sorry, but we won't be able to use Tabs component from that package 😣

We can either create our own basic Tab component or choose another design, like this:

What do you think? I think it looks cleaner and we already have the Toggle component (it could be much smaller though...)

screenshot 2018-10-10 at 20 50 53

ganesharulanantham:
@goenning got it. will first try to implement our own tab implementation. It fits for all pages as text area   and then will try toggle. Toggle also sounds like a good idea.

goenning:
I kind of liked the Toggle more than Tabs, mostly because it uses less space and there's nothing between the input and the textarea. I'm afraid that the "Write" + "Preview" floating text might make users confused.

goenning:
Hmm, that’s actually a good point. Well, give it a try and let’s see how it looks like. Thst one one is not frequently used, as it’s only visible to staff users. I’d prefer one that fits better on the home/comment screen.

Issue #307, opened 2018-03-11
Cache Gravatar response


goenning:
**Problem:** Gravatar pictures are proxied through Fider to keep user’s emails private. Right now, every request to Fider Gravatar proxy will send a new request to Gravatar, which can add some overhead to the servers if there are many users online.

**Solution:** Gravatar imagens should be cached for a few hours/days in-memory at Fider server.

A `map[string][]byte` is a good candidate to cache the images.

nylen:
@goenning: an alternative would be to have the browser request a Gravatar directly.  This is what all WP sites do, and it's [the expected use case](https://en.gravatar.com/site/implement/images/) for other implementations as well.  Would this be an acceptable solution?

goenning:
We used to do that, but we got an interesting bug report that it’s easy to reverse engineer the md5 code and “steal” the email from users, which is a valid point!

Since then we’ve been using Fider as a proxy to Gravatar.

cameronmoreau:
It looks like this issue has been open for a while so I'll go ahead and take a stab!

PR #564, opened 2018-10-10
Cache Gravatar response

cameronmoreau:
**Issue:**
Cache Gravatar response #307

**Solution:**
The images#Avatar method will check if the Gavatar url is stored in a cache, returning the stored byte array if hit or caching for 24 hours if a miss.

Originally, I thought to add a simple `map[string][]byte` to `storage/inmemory/images.go`, but this does not work as `WebSetup` is called every request and will not retain a cache.

Since an  in-memory cache isn't set up yet, I went ahead and added [go-cache library](https://github.com/patrickmn/go-cache) and made a cache accessible via `Engine`. I thought that this library had a great set of features, including item expiration and cleanup.

**Note:**
I had an issue adding the library to Gopkg.lock. Running `dep ensure` added many other keys to the lock file such as `digest`. Maybe this is a dep version mismatch? This may need to be looked at.

Also, I wasn't quite sure on the best testing strategy for this feature.

Feel free to leave review for edits!

**Useful Logs**
```
INFO [2018-10-10T14:53:30-05:00] [WEB] [SYi6BSNL2rlXMgCkj7LASILltEa67MDg] GET http://dev.assets-fider.io:3000/avatars/50/1/Cameron%!M(MISSING)oreau started for 127.0.0.1
...
DEBUG [2018-10-10T14:53:30-05:00] [WEB] [SYi6BSNL2rlXMgCkj7LASILltEa67MDg] Requesting gravatar: https://www.gravatar.com/avatar/XXX?s=50&d=404
INFO [2018-10-10T14:53:30-05:00] [WEB] [SYi6BSNL2rlXMgCkj7LASILltEa67MDg] GET http://dev.assets-fider.io:3000/avatars/50/1/Cameron%!M(MISSING)oreau finished in 391ms (committed)
INFO [2018-10-10T14:53:32-05:00] [WEB] [NxivsrIiFGPsGfczYdLROrx7Wxmrn2ZL] GET http://dev.assets-fider.io:3000/avatars/50/1/Cameron%!M(MISSING)oreau started for 127.0.0.1
...
DEBUG [2018-10-10T14:53:32-05:00] [WEB] [NxivsrIiFGPsGfczYdLROrx7Wxmrn2ZL] Gavatar found in cache: gavatar:https://www.gravatar.com/avatar/XXX?s=50&d=404
INFO [2018-10-10T14:53:32-05:00] [WEB] [NxivsrIiFGPsGfczYdLROrx7Wxmrn2ZL] GET http://dev.assets-fider.io:3000/avatars/50/1/Cameron%!M(MISSING)oreau finished in 57ms (committed)
```


goenning:
I had a quickly look and it seems great, but I’ll review more carefully later this week. I’m not a dep expert, but i believe dep ensure will try to update the dependencies, which we don’t want to.

Try dep ensure -add GitHub.com/...

You’ll also need to commit that library to vendor package, like we do with all the other libraries, otherwise the build will not go green.

Thanks for the PR! 😀

cameronmoreau:
Ahh, didn't even notice that vendor was checked-in, thanks!

After running `dep ensure -add github.com/patrickmn/go-cache` the lock file was updated with the extra keys. It looks like dep tries to add mage to the lock file and pulled it into the vendor folder aswell.

This recent commit only adds the library in the vendor folder, and not the newly generated lockfile

goenning:
I was using an older version of Dep, my bad. I have updated to dep 0.5.0 (latest release) and pushed to master. So you should now be able to pull in and run `dep ensure`, that'll change your lock file accordingly without changing every other entry. 

I have also ignored mage package, so that would be downloaded into vendor anymore.

Issue #569, opened 2018-10-12
missing Date and Message-Id headers on SMTP email delivery


goenning:
**Problem:** Spam Assassin doesn't like emails without Date and Message-Id headers, which is exactly what Fider currently does.

**Solution:** We need to set both headers for every email sent via SMTP. As we don't have any requirement for ingesting e-mails, we don't really care about the Message-Id that we use. So for now we could generate a random string with a reasonable length (128?!).

For reference:
- https://wiki.apache.org/spamassassin/Rules/MISSING_MID
- https://wiki.apache.org/spamassassin/Rules/MISSING_DATE

lukechampine:
I can take a stab at this if you point me towards which file(s) it would touch. :)

goenning:
Nice! You'll only need to change `fider/app/pkg/email/smtp/smtp.go` and update the unit tests (smpt_test.go) accordingly. Thanks! :)

goenning:
@lukechampine have you started working on it? @my3 just sent a PR for it

lukechampine:
ah, too slow! I won't bother sending a duplicate PR, then -- although I would suggest using `crypto/rand` + `hex.EncodeToString` instead of importing a `uuid` package.

goenning:
@lukechampine how is that better than using `math/rand` for a random string generator? I understand that crypto/rand is used for operations that require a cryptographically secure random number. But for our use case, would that provide any benefit? What we need here is a unique string generator (like `uuid`) or a long random string generator. We have this function that I was hoping we could use, instead of importing another external package.

https://github.com/getfider/fider/blob/ae9f3d48506a5e9c8a706a2d630a8102661c15ed/app/pkg/rand/random.go#L12-L26

But if there's a reason crypto/rand+hex would produce better results, i'm all in for using it.

lukechampine:
One issue with `math/rand` is seeding. By default, it will be seeded with 0, so every user will be generating the same numbers. Seeding with the current time will probably alleviate this, but...why risk it?

`hex` is just for convenience, so you can write:
```go
buf := make([]byte, 16)
rand.Read(buf)
return hex.EncodeToString(buf)
```

my3:
I was about to say that `uuid` also uses `crypto`. Thought uuid generation will most likely be needed in other places as well and so included that package. If we decide to go with `crypto`, I can make those changes.

my3:
@lukechampine In case you already have code with `crypto`, you can raise a PR and we can close current one.

[...]

PR #573, opened 2018-10-13
add Date and Message-ID to mail

my3:
**Issue:**   #569


app/pkg/email/smtp/smtp.go
```diff
@@ -157,3 +168,15 @@ func sendMail(localName, serverAddress string, a gosmtp.Auth, from string, to []
 	}
 	return c.Quit()
 }
+
+func generateMessageID() (string, error) {
+	timestamp := strconv.FormatInt(time.Now().UTC().UnixNano(), 10)
+	unq := uuid.New()
+	randStr := unq.String()
+	host, err := os.Hostname()
```
goenning: We shouldn't use os.Hostname here, the hostname is usually completely different than the real domain name. There's a `localname` variable that we could use instead.

app/pkg/email/smtp/smtp.go
```diff
@@ -157,3 +168,15 @@ func sendMail(localName, serverAddress string, a gosmtp.Auth, from string, to []
 	}
 	return c.Quit()
 }
+
+func generateMessageID() (string, error) {
+	timestamp := strconv.FormatInt(time.Now().UTC().UnixNano(), 10)
+	unq := uuid.New()
```
goenning: I'd prefer to use `rand.String(32)` that is already available on Fider. This random string + timestamp is unique enough for this operation. By doing so, we won't need the external `uuid` package.

app/pkg/email/smtp/smtp.go
```diff
@@ -157,3 +163,12 @@ func sendMail(localName, serverAddress string, a gosmtp.Auth, from string, to []
 	}
 	return c.Quit()
 }
+
+func generateMessageID(localName string) string {
+	timestamp := strconv.FormatInt(time.Now().UTC().UnixNano(), 10)
+	buf := make([]byte, 16)
+	rand.Read(buf)
```
goenning: `rand.Read` might throw an error, that's very unlikely to happen, but if does, we should panic. We have a top `recover` goroutine that handles panics.

```suggestion
	_, err := rand.Read(buf)
	if err != nil {
	  panic(err)
	}
```

Gopkg.lock
```diff
@@ -173,6 +173,7 @@
     "github.com/dgrijalva/jwt-go",
     "github.com/goenning/letteravatar",
     "github.com/goenning/sqlcertcache",
+    "github.com/google/uuid",
```
goenning: Just revert all changes to Gopkg.lock

Gopkg.toml
```diff
@@ -37,3 +37,4 @@ ignored = ["github.com/magefile/mage/mg","github.com/magefile/mage/sh"]
   branch = "master"
   name = "github.com/goenning/sqlcertcache"
 
+
```
goenning: Just revert all changes to Gopkg.toml

Gopkg.lock
my3: ah.. thought I reverted all changes. will update and squash commits or create a new PR with one commit

goenning:
Thank you so much for this! 🎉

Issue #576, opened 2018-10-13
Allow users to mention/tag other users


goenning:
Source: https://feedback.fider.io/posts/82/allow-users-to-mention-tag-other-users

The goal is to implement exactly what GitHub does when you start typing `@`, where it suggests a list of users to "tag". Whenever tagged, we should send a notification (both e-mail and in-site) about they being mention on that post.

The challenge we have here is that Fider doesn't have username, just Name. More importantly, names can have empty spaces and special character (like @#*!). We could think about a different way of doing so, but using @ is becoming well known pattern, it's something people are getting used to it.

ihatetothink:
I'm thinking about taking this task, would it be for comments or also when creating new suggestions?

I believe we should match "@<not spaces>*<cursor>" does this sound right? 
Also we need a new endpoint that returns all the users right?

goenning:
I'd say just for comments, at least for now. 
I think we can start simple and improve afterwards. What you suggested is good enough. I'd just say that hitting `Enter` should also select the first suggestion.

Yes, we'll need an endpoint where the input is the search term and the result is an array of users with its ID and Name.

ihatetothink:
About the frontend part, can I use some react component library? I'm not that fluent with react

goenning:
I'd rather avoid external libraries. If you have the time, it might be a good moment to level up your react skills with this task. I reckon that is not easy though, but I can help you if you need. 

I'm also not sure how I would implement this, so I'd start with reading other codebases/blogs to see how other people implement such feature.

ihatetothink:
Ok, doing that. 

Also fork is here: https://github.com/ihatetothink/fider for now I have the endpoint

goenning:
You could open a PR, prefix it with [WIP] and ping me everytime you’ve finished part of it.

PR #579, opened 2018-10-17
[WIP] Allow other users to mention/tag other users

ihatetothink:
**Issue:**  
https://github.com/getfider/fider/issues/576

This pull request will include frontend and backend functionality to allow tag/mention suggestions to the user while writing a comment and/or when creating a new suggestion.

We could also add here the notification of tagged users by email.

What has been done:
- Backend endpoint to list users starting with a prefix

Missing:
- Frontend Part
- Sending email to mentioned users

About the FE part:
- Looking around I found quite a few React libraries, if we implement that from scratch I'd say it's not a trivial task. Most of the libraries I've found create div elements and position them absolutely in the cursor position. (I'm still looking at alternatives)


ihatetothink:
Yes I did read it! Then using react-input-trigger is ok?

app/storage/postgres/user.go
```diff
@@ -346,3 +345,22 @@ func (s *UserStorage) GetByAPIKey(apiKey string) (*models.User, error) {
 	}
 	return user, nil
 }
+
+func (s *UserStorage) FindLike(query string) ([]*models.User, error) {
+	var users []*dbUser
+	err := s.trx.Select(&users, `
+		SELECT id, name, email, tenant_id, role, status 
+		FROM users 
+		WHERE name like $1 || '%'`, query)
```
goenning: Two other filters we are missing here:
- tenant_id, so that it doesn't list users from another tenant.
- status, so that it doesn't not list deleted users.

Have a look on `GetAll()` function, it does a filter on both fields that we also need to filter here.

ihatetothink:
I'm not sure how to integrate that library:
- Library wants to have a text area child
- We have this TextArea react object on top of all textareas

My plan is to integrate it directly in our TextArea class, exposing 
- some flag/parameter enable it or disable triggering
- some parameter that can receive this triggering events

I might end up not integrating it and just taking the parts of code that we need, as the library is not too complex and is quite rigid. 

Is it a problem if all textboxes end up being one level deeper? as they will be wrapped in that InputTrigger class

Any comments would be appreciated as I do not find my solution really correct. Thanks :)

goenning:
I think we a new props on TextArea component, called `allowMentions` maybe? When this is set to true, we can wrap and enhance the textarea with this feature. I'm not sure if we'd need to bubble up any event to parent component when a user is mentioned.

To be honest, I don't have a better idea. I'd probably do the same as you mentioned.  I don't think having another level on the tree is bad. We can always refactor later if needed. And I also like the idea of pulling in parts of that library instead of the full library 😄 

We might need to think about the format of the mention that is sent to server. Fider supports multiple users to have same name, so the ID probably a must have.

Option 1:

If the user types: `Hello @Agusti Bai, how are you?`
Should we send to server something like: `Hello __user_18294__, how are you?`? This could then be parsed by the server, send an email to user with id 18294 and then save `Hello **@Agusti Bai**, how are you?` into database so it can be easily rendered later on.

Option 2:

If the user types: `Hello @Agusti Bai, how are you?` Send the raw text + a list of ID of the users that needs to be mentioned, in this case, just a { `content: "Hello @Agusti Bai, how are you?", mentions: [ 18294 ] }`

Option 3:

??? :)

goenning:
I see that your making a great progress on it! Just let me know whenever you want me to review what you’ve done so far.

ihatetothink:
I believe now it's a good moment to review what's been done. The full mention suggestion component is there.

What is missing:
- Add grammar to allow mentioning. _[<USER ID>] sounds good to me_
    - Display the user name when viewing mention. 
    - Interpret grammar to send email/notification when mentioning. _I'd do this when creating only)_

What I'm not sure should be part of this issue:     
- Create email to be sent

What is missing and I'd do outside this PR (to keep it smallish and following issue description):
- Add post mentioning. (Right now is only implemented for comments to reduce the scope a bit)
- Add on email update comment/post with a mention (will require storing historic data)

ihatetothink:
I have added the logic to send emails and notifications when a user is mentioned. (can be triggered by writing a comment containing [id], suggestion is not rigged yet)

missing:
- Modify the comment view so that it shows usernames instead of [id] when writing the comment, allowing to delete the mention. (I'm not so sure how to tackle that)
- Modify the mention suggestion so that it adds [id] instead of the name itself

For the comment view we can do two things
1: Store comments with the username itself
2: Interpolate it when requesting for the post comments

Not sure which one is better

Also we should add some setting to opt in/out of those communications I believe right?

goenning:
Some minor things I found when reviewing this.

- move @types/textarea-caret to devDependencies
- add textarea-caret to the vendor array on webpack.config.js. This package won't change very often, so it's better to vendor it.
- Send the user array into SuggestionBox so we can show the user <Gravatar />. When displaying the user name, use <UserName /> component
- Allow user to use mouse to select suggestion box item :)
- I think the backend search should be case insensitive and search any part of the name. 'oma' should match 'Tomas'. What do you think? We might need to limit number of rows to 10 (?)
- When the search results return an empty array, if the user continue to type something, it shouldn't query the database anymore because it won't find anything.
- you can remove `import { number } from "prop-types";`

I really liked how you organized the components, very clean and reusable! Great job! 👍 

This is going to be a bigger PR than I imagined. If you agree, I'd suggest we split it into two PR.

For this PR, we could:

- change MentionableTextAreaState to have an array of mentioned users. Right now it only accepts one, but in reality it would be more than one.
- The array of mentioned users could be something like: `mentions: Array<{ user: User, start: number, end: number }>`
- Revert CommentInput so that it uses TextArea again, at least until the whole feature is ready.
- Allow user to delete a mention. This is going to be tricky. Facebook uses a div with contentEditale, so they insert HTML into the "text area". So when you click backspace, it deletes the whole Mention block, not just a character.

On the second PR we could work on the how to send this information to server, parse and notify users, and then switch the UI component 😄 

What do you think?

Thank you!

goenning:
I just though about something, don't get angry with me, please! 😃

We recently introduced a code split setup on Fider, so we can now load components asynchronously.

This mention component is way more complex than I first though, there are some very complex edge cases (like deleting). We could introduce an external library using React Suspense. That way only users that click on "Leave a comment" would download this external library, without affecting the rest of the platform. Any thoughts on it?

I'm really sorry to mention it only now, but it just came to my mind 😢

goenning:
Yes, exactly. My main concern with external libraries is that they are usually bigger than necessary and thus degrade performance. So I usually try to implement my own lean components. But now that we have Code Splitting, that's less of a problem! That doesn't mean we will add tons of external libraries, but for some complex scenarios, I think it's worth to consider them. 

Mentions has proved to be complex enough and the package you mentioned seem to implement everything what we need. As long as we use Suspense+Lazy, I think we could use that. What do you think about it?

ihatetothink:
Let's do that, I've checked that library code and seems like corner cases don't end in deleting mentions, different browsers have some quirks as well. I'll get into it early next week :)

Issue #532, opened 2018-09-12
SMTP office365 email not sent


davidbonachera:
Hi there,

So I tried to use my office365 smtp by setting off my parameter like this :  
```docker
- EMAIL_SMTP_HOST=smtp.office365.com
- EMAIL_SMTP_PORT=587
- EMAIL_SMTP_USERNAME=email@aaa.com
- EMAIL_SMTP_PASSWORD=Ipassword
```

But usually it requires TLS encryption, so I tried to add an encryption method : 
```
- EMAIL_SMTP_ENCRYPTION=tls
```

But still, no e-mail are being sent.

There's no SMTP error in a dev environment inside the docker log, any idea how to debug this situation ? (i'm not an expert in go)

Tks in advance,

goenning:
After some research I found out that Go stdlib `net/smtp`doesn’t support Office365 out of the box. We’ll need use an external package or implement it ourselves. 

For reference: https://gist.github.com/homme/22b457eb054a07e7b2fb

chansuke:
I'd like to pick this up:-)

goenning:
@chansuke you’re more than welcome! 😀

goenning:
I forgot to mention that if you need anything, don't hesitate to ask! I can't promise an answer to all questions, but will do my best!

chansuke:
@goenning thanks!!!

goenning:
This is now possible thanks to @oxynux contribution! 😄

davidbonachera:
Thanks I will wait for the 0.17 release !

PR #584, opened 2018-10-19
Allow support of Office 365

oxynux:
Issue: #532

This allow fider user's to set `EMAIL_SMTP_AUTH_LOGIN=1` in `.env` file, for support Office 365 SMTP (smtp.office365.com) and others SMTP server who use "LOGIN Simple Authentication".

Office 365 SMTP use "LOGIN Simple Authentication" defined by this Internet Draft: [draft-murchison-sasl-login-00.txt](https://www.ietf.org/archive/id/draft-murchison-sasl-login-00.txt).


goenning:
That looks great, thank you @oxynux 

I have a suggestion if you don't mind. I'd like to avoid the extra environment variable and I believe it's possible.

The gotsmtp.ServerInfo has an `auth` property that is an array of all supported Auth mechanisms of the SMTP server. We can use that to decide whether we use PlainAuth or LoginAuth. What do you think? We might end up having a third Auth struct that decides which of those other two Auth methods to use.

By avoiding the extra parameter we make it flexible and easier for the end user :)

oxynux:
@goenning It should be ok 🙂

goenning:
@oxynux that's great! I didn't know that Go smtp had support for CRAMMD5Auth. So this PR will fix a bug and introduce a feature, great job 👍 ! Looking forward to merge this one.

I have my last two request, hope you don't get angry at me 😃 . 

We should remove that duplicate code of initialising the Auth's twice + calling `Start(server)` and `Next(fromServer, more)` twice as well. 

If we extract the logic to find the correct auth into a new function, like. `func (a *agnosticAuth) FindAuth(server *gosmtp.ServerInfo) gosmtp.Auth` that can return either an instance of a gosmtp.Auth or nil. You can then use call this function on Start and store the result into agnosticAuth (instead of saving just a string). This instance can then be reused on Next.

We also need to cover AgnosticAuth with simple unit tests. The ones I can think about is: 

1. It should choose the correct Auth method based on the auth string array;
2. It should return an error if we don't support any of the auth string array;

oxynux:
@goenning Sorry for the response delay.

I agree it's sounds good to remove duplicate code, however I don't really understand how to "extract the logic", maybe when I will put my mind to it I'll understand better 😅.

I'll try to finish this this weekend.

goenning:
That's ok, don't worry! If you need something I'll be online most of the weekend so just ping me :)

oxynux:
`func (a *agnosticAuth) FindAuth(server *gosmtp.ServerInfo) gosmtp.Auth` is now implemented 😀
@goenning I'm not familiar with fider unit test suite can you give me an exemple ?

goenning:
Awesome! That's exactly what I was thinking about 😄

As for the test, you can create a `auth_test.go` file and paste this as an example:

```
package smtp_test

import (
	gosmtp "net/smtp"
	"testing"

	. "github.com/getfider/fider/app/pkg/assert"
	"github.com/getfider/fider/app/pkg/email/smtp"
)

func TestAgnosticAuth_Login(t *testing.T) {
	RegisterT(t)

	auth := smtp.AgnosticAuth("", "jon", "s3cr3t", "test.fider.io")
	proto, bytes, err := auth.Start(&gosmtp.ServerInfo{
		Auth: []string{"LOGIN"},
	})

	Expect(err).IsNil()
	Expect(proto).Equals("LOGIN")
	Expect(bytes).Equals([]byte("jon"))

	bytes, err = auth.Next([]byte("Username:"), true)
	Expect(err).IsNil()
	Expect(bytes).Equals([]byte("jon"))

	bytes, err = auth.Next([]byte("Password:"), true)
	Expect(err).IsNil()
	Expect(bytes).Equals([]byte("s3cr3t"))
}

func TestAgnosticAuth_NoMatchingAuth(t *testing.T) {
	RegisterT(t)

	...
}
```

The second test you can start like the first, but set `Auth: []string{"FAKE-MD5"}`. This is kind of mocking a SMTP server that only supports an auth mechanism that we don't implement. The result should be that `smtp.AgnosticAuth` returns an error. :)

oxynux:
@goenning It's finally done 🎉
Thanks for the test exemple, implementing `TestAgnosticAuth_NoMatchingAuth(t *testing.T)` helped me fix `Start` method of `AgnosticAuth`.

goenning:
Thank you @oxynux! It's a great contribution 🎉

Issue #580, opened 2018-10-18
Burger menu is incorrectly positioned


goenning:
As an Admin/Collaborator, go to Site Settings. Resize the browser to the smallest windows size possible. The menu will be collapsed into a burger menu, but it won't be aligned with page title as it should.

<img width="499" alt="screenshot 2018-10-18 at 07 24 36" src="https://user-images.githubusercontent.com/94755/47135190-3a0b7180-d2a7-11e8-813a-e4ffcdf5fbec.png">

damcclean:
I'll see what I can do :)

peterver:
@goenning I'll take a look at this one over the weekend :D

PR #587, opened 2018-10-20
[fider/580] Burger menu is incorrectly positioned

peterver:
**Issue:**  #580 

Correctly position the burger menu collapser at the top right corner of the container that it is in.


goenning:
This is great, thank you! 😄

Issue #581, opened 2018-10-18
Invalid Posts should return a 404 and not a 500


goenning:
**Problem**: at the moment, if we try to navigate to a `View Post` page that doesn't not exist, we'll receive a `500 Shoot! Well, this is unexpected…`, but we should actually receive a `404 Not Found`

**Note:** The same happens to the API, if we try to call `/api/v1/posts/abc`, the response will be 500, but should be 404.

peterver:
I'll take a look at this one over the weekend !

PR #589, opened 2018-10-20
[fider/581] Invalid posts should return a 404 and not a 500

peterver:
**Issue:**  #581 

- Adjust the handlers to send back a 404 if the `ParamAsInt("number")` validation fails.
- Add an endpoint to retrieve a post (Rest)

### DIFF
--- a/app/cmd/routes.go
+++ b/app/cmd/routes.go
@@ -159,7 +159,7 @@ func routes(r *web.Engine) *web.Engine {
 		api.Use(middlewares.IsAuthenticated())
 
 		api.Post("/api/v1/posts", apiv1.CreatePost())
-                api.Get("/api/v1/posts/:number", apiv1.GetPost())
+		api.Get("/api/v1/posts/:number", apiv1.GetPost())
 		api.Post("/api/v1/posts/:number/comments", apiv1.PostComment())
 		api.Put("/api/v1/posts/:number/comments/:id", apiv1.UpdateComment())
 		api.Delete("/api/v1/posts/:number/comments/:id", apiv1.DeleteComment())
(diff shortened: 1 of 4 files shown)


app/cmd/routes.go
```diff
@@ -159,6 +159,7 @@ func routes(r *web.Engine) *web.Engine {
 		api.Use(middlewares.IsAuthenticated())
 
 		api.Post("/api/v1/posts", apiv1.CreatePost())
+                api.Get("/api/v1/posts/:number", apiv1.GetPost())
```
goenning: Wow, how did I miss this API? Thank you!

Few comments:

1. We can move it before the IsAuthenticated middleware, right next to  `/api/v1/posts`
2. Align it with the rest of the code.

app/handlers/apiv1/post.go
```diff
@@ -52,6 +52,23 @@ func CreatePost() web.HandlerFunc {
 	}
 }
 
+// GetPost retrieves the existing post by id
```
goenning: id -> number

// GetPost retrieves the existing post by number

peterver:
@goenning will take a look at the roc tomorrow :) !

peterver:
@goenning changes are done :)

peterver:
@goenning I also added a test for the new GetPost route

goenning:
@peterver  Add a `Expect(query.String("title")).Equals("My First Post")` assertion to that query :)

goenning:
@peterver one more thing, you need to move the `api.Get("/api/v1/posts/:number", apiv1.GetPost())` before the `api.Use(middlewares.IsAuthenticated())`. Posts are public and without that only authenticated users would be able to use that API.

peterver:
there ya go @goenning :)

goenning:
Approved! But looks like CircleCI didn't run the build :(

peterver:
@goenning probably stuck in a queue somewhere :/. Anyway to retrigger a circleci build or let's just way for the queue to do its thing ?

goenning:
I don't see any queue on CircleCI. Can you change the title or description on that test and push another commit? That should force a new build.

peterver:
@goenning There we go, now circle ci is rollin :)

goenning:
That's merged, thank you!

I'll prepare and cut the 0.16.0 release now! 🎆

Issue #588, opened 2018-10-20
 Implement Code Splitting ⚡️


goenning:
We currently bundle every JavaScript and CSS into their own single bundle. We have a vendor bundle because third party packages change way less often than our own code, so that's a problem solved.

The problem we have now is that a small change on a specific component results into a completely new bundle which has to be downloaded by all users.

We should really look into Code Splitting and how that can help us push smaller javascript files to users. We could also think about using `optimization.splitChunks`.

My idea is to split based on our Page Components that are described on [router.tsx](https://github.com/getfider/fider/blob/master/public/router.tsx). Every Page Components would result into a separate bundle that is downloaded only when necessary.

Tasks:
- [x] Split JavaScript by Routes (PR: #596)
- [x] Fix CSS Warnings (PR: #619)
  - Reminder 1: we have multiple `import "./TagForm.scss"`. This css file should only be imported by TagForm. This was probably a copy-paste mistake.
  - Reminder 2: `npx webpack -p --profile --json > compilation-stats.json` generates a very useful internal view of modules. This might help us resolve this issue.
- [x] enable esModuleInterop and fix everything that breaks after that. Jest tests should be warning-free after this. (PR #600)
- [X] Move common components (`public/components/common`) to its own JS and CSS bundle. These components don't change that often, thus avoiding users having to download this over and over again. #636
- [x] Webpack allows us to give name to chunks, like `import(/* webpackChunkName: "ShowPost.page" */ "@fider/pages/ShowPost/ShowPost.page");`. We can use this to preload this chunk as shown here: https://developers.google.com/web/updates/2016/03/link-rel-preload (PR: #619)
- [x] Replace React-Loadable with React.lazy. (PR: #646)

Resources:
- https://webpack.js.org/guides/code-splitting/
- https://reactjs.org/docs/code-splitting.html
- https://webpack.js.org/plugins/split-chunks-plugin/

PR #600, opened 2018-10-26
Esinterop enable

cfilby:
**Issue:**  Add esInterop and change import styles from `import * as X` to `import X`. Also remove some unused imports. Subtask of #588.


goenning:
Awesome! This change even reduced the main bundle by 19 KiB, that's unexpected, but highly appreciated! 😄

PR #601, opened 2018-10-26
Delete old logo upload whenever a user updates his current logo

### LINKED ISSUE #479: remove unused uploads on startup
**Problem:** Every time a user uploads a new Logo, Fider will create a new record on `uploads` table and update the `tenants` table field `logo_id` to the new `id` generated from the `uploads` table. Every new page will use the new logo instead of the previous one. The problem is that we now have old logos on the `uploads` table that are not used anymore.

**Solution:** The solution I can think about is to remove all orphaned `uploads` records on startup. We could also remove when we upload new logos, but that means we'd need to do apply this change to multiple places in our codebase, not to mention any other future file upload would also require this.

chrisprobst:
**Issue:**  #479

Whenever a tenant updates the logo, the current logo upload row is deleted.

### DIFF
--- a/app/storage/postgres/tenant.go
+++ b/app/storage/postgres/tenant.go
@@ -205,6 +205,14 @@ func (s *TenantStorage) UpdateSettings(settings *models.UpdateTenantSettings) er
 			newLogoID.Scan(uploadID)
 		}
 
+		// Try to find out if there is an upload for the logo id of the tenant
+		oldLogoExists, err := s.trx.Exists("SELECT content_type, size, file FROM uploads WHERE id = $1", s.current.LogoID)
+		if err != nil {
+			return errors.Wrap(err, "failed to get old tenant logo upload")
+		}
+
+		oldLogoID := s.current.LogoID
+
 		// Update current tenant logo to either new ID or null
 		query := "UPDATE tenants SET logo_id = $1 WHERE id = $2"
 		_, err = s.trx.Execute(query, newLogoID, s.current.ID)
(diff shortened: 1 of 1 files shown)


app/storage/postgres/tenant.go
```diff
@@ -205,6 +205,14 @@ func (s *TenantStorage) UpdateSettings(settings *models.UpdateTenantSettings) er
 			newLogoID.Scan(uploadID)
 		}
 
+		// Try to find out if there is an upload for the logo id of the tenant
+		oldLogoExists, err := s.trx.Exists("SELECT content_type, size, file FROM uploads WHERE id = $1", s.current.LogoID)
```
goenning:
```suggestion
		oldLogoExists, err := s.trx.Exists("SELECT id FROM uploads WHERE id = $1 AND tenant_id = $2", s.current.LogoID, s.current.ID)
```

If we don't need all those fields, better return just one. We also need to make sure tenant_id is always filtered on all SQL commands.

app/storage/postgres/tenant.go
```diff
@@ -214,6 +222,12 @@ func (s *TenantStorage) UpdateSettings(settings *models.UpdateTenantSettings) er
 
 		// Update reference to new logo ID
 		s.current.LogoID = int(newLogoID.Int64)
+
+		if oldLogoExists {
+			if _, err = s.trx.Execute(`DELETE FROM uploads WHERE id = $1`, oldLogoID); err != nil {
```
goenning:
```suggestion
			if _, err = s.trx.Execute(`DELETE FROM uploads WHERE id = $1 AND tenant_id = $2`, oldLogoID, s.current.ID); err != nil {
```

Same as above, we always filter by tenant_id.

PR #602, opened 2018-10-28
Unit tests: UTC changed to local time

Mygeen:
**Issue:** Unit test was using UTC comparing to local time. Therefore, the test failed in time zone UTC +1.

I edited the test values of the failing unit test to use local time instead of UTC.


goenning:
Nice one! Thanks!

Issue #604, opened 2018-10-28
Windows dev environment common issues and python as required tool


renatoch:
While setting up environment for working on windows, got 2 problems that I found worthwhile including in common issues documentation.
Also, found that node-sass uses python 2.7.X while installing, so included that in required tools.

goenning:
Are these the only issues you had on Windows? Is Fider running as expected?

Issue #603, opened 2018-10-28
Error during setup for developing on Windows 10


renatoch:
> Cannot download "https://github.com/sass/node-sass/releases/download/v4.9.3/win32-x64-67_binding.node":
> HTTP error 404 Not Found

Tried changing to 4.9.4 on package.json, but not sure that's the correct way to fix this.

goenning:
Hey, are you using Node.js 11? If so, then that’s the issue. Because node-saas does not yet provide binding for node 11 (it’s a very recent release).

That’s why you beed Python to compile it yourself.

PR #605, opened 2018-10-28
Windows dev environment common issues and python as required tool

renatoch:
**Issue:**  #604 

- Includes 2 problems in common issues documentation found while setting up on windows
- node-sass uses python 2.7.X while installing - included that in required tools


CONTRIBUTING.md
```diff
@@ -22,6 +22,7 @@ If you know these technologies or would like to learn them, lucky you! This is t
 | Go 1.11+ | https://golang.org/ | To compile server side code |
 | Node.js 10+ | https://nodejs.org/ or run `nvm use` if you have nvm installed | To bundle client side code |
 | Docker | https://www.docker.com/ | To start local PostgreSQL instances |
+| Python 2.7.X | https://www.python.org/downloads/ | To automatically build node-sass npm package during its installation |
```
goenning: If my assumption on #603 is correct, we should drop this requirement. It's only a requirement for very specific and rare edge cases. In 99% of the cases Python is not required.

CONTRIBUTING.md
```diff
@@ -57,3 +58,15 @@ This happens when godotenv was not (or incorrectly) installed. Install it by run
 #### 3. mage watch throws 'too many open files' error
 
 macOS has a small limit on how many files can be open at the same time. This limit is usually OK for most users, but developers tools usuage require a larger limit. [Learn how to resolve this](https://www.macobserver.com/tips/deep-dive/evade-macos-many-open-files-error-pushing-limits/).
+
+#### 4. mage isn't found even after installing (Windows)
```
goenning: This is a good one! But it's not specific to Windows. If we make the text OS-agnostic, we can add this.

Issue #575, opened 2018-10-13
Translate Fider to other languages


goenning:
Source: https://feedback.fider.io/posts/8/translate-fider-to-other-languages-internationalization-i18n

Fider is currently available only on English and our the long term goal is to internationalise it. Being an open source project, we hope to engage with the global community to help us translate it to every existing language ❤

But before we get there, there's a long path we need to trail to make our strings translatable. The goal of this issue is to track the Plan and Development of this feature. Join us with ideas, feedback and code contributions!

Definitions

1) The language is defined per tenant/site and is not based on current browser language (Accept-Language HTTP Header);
2) The user submitted content WILL NOT be translated;
3) If a string is not translated, it'll default to English. So it's perfectly possible for some sites to have a mixed Non-English/English text for languages that are not 100% translated; We should show the percentage of completion for each language when user is selecting a language;
4) The browser should only download the strings for the required language.
5) The strings that are stored in `JSON/XML/TXT/whatever` must be available to both Go and React.

Questions:

1) Which languages to start with? I (@goenning) can translate it to English and Brazilian Portuguese
2) How can we support RTL languages? Is it hard?
3) Which platform to use? GitHub? Crowdin? What other?
4) Which format to store the strings? JSON?
5) How do we introduce breaking changes to existing strings? Say we have: `{ key: "NOT_FOUND_MESSAGE", value: "The page is not found." }` and this has been translated to every language. What if we change text to `This page does not exist.`? How do "deprecate other" strings that have already been translated.
6) Any Go+React package that can help us on this? It doesn't seem to be hard to implement, but other packages might give us some inspiration.
7) Do we need different validation per language? E.g.: Latin alphabet we should validate that titles must have at least 10 chars. But on Chinese/Japanese, maybe 2 or 3 is enough. (ref: #614)

If anyone has experience with a multilingual platform, please share your thoughts and suggestions!

lkraider:
Hello,

I would recommend centering the message interchange format in `gettext`, it is the de-facto standard, and multiple cloud based translation collaboration tools work off of that.

For the specific language tools, then trace the ones that can convert bidirectionally from their own format, if any, or work directly with gettext.

For JS React (that have the mentioned features) I've looked into:

i18next:
https://www.i18next.com/
https://github.com/i18next/i18next-gettext-converter

Lingui:
https://lingui.js.org
https://lingui.js.org/ref/conf.html#po

Sven65:
I am available to do some translation into Swedish

[...]

PR #614, opened 2018-11-02
title validate support Chinese and Japanese

fwchen:
**Issue:**  
When I type in Chinese the system cannot be used as easy as in English, just like when I edit the title. 

I just remove `|| len(strings.Split(input.Model.Title, " ")) < 3` because I think it is unnecessary when support other language.

### DIFF
--- a/app/actions/post.go
+++ b/app/actions/post.go
@@ -33,7 +33,7 @@ func (input *CreateNewPost) Validate(user *models.User, services *app.Services) 
 
 	if input.Model.Title == "" {
 		result.AddFieldFailure("title", "Title is required.")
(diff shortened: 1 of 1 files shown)


goenning:
Hmm, interesting. What about the first validation `len(input.Model.Title) < 10`? Isn't it also irrelevant on Japanese/Chinese? I guess you can have a very descriptive title with very few chars, or am I wrong? It makes me thing that #575 will be more than just translations, it's also about settings per language.

fwchen:
emmm... I think `len(input.Model.Title) < 10` is enough. It can almost support all the languages.

ok, I will fix the test a few days later.(triping, only on chromebook)

goenning:
Merged, thank you!

PR #619, opened 2018-11-03
Add Webpack Chunk Names and Fix CSS Warnings

cfilby:
**Issue:**  #588 

Adds named webpack chunks to the Async loader and fixes the CSS Warnings by reordering imports in Home/index.ts. 

The duplicated TagForm CSS imports have been removed as well by the previous code splitting branch.


goenning:
The chunk name is great, but we still need to add the preload tag to get the performance benefit.

`<link rel="preload" href="path/to/Home.page.8172371283" as="script">`

My idea was to include a `ChunkName` field to `web.Props`. Every occurrence of `return  c.Page(...)` should also set the ChunkName to be preloaded. On `web.Rendered` we can then use this field to get the full bundle name (like we do with main.js and vendor.js) and render it on the HTML template.

By the way: maybe this change that fixed the CSS warning on `mage watch`, but there are still CSS warnings on `mage build`. Don't need to be included on this PR though...

Thanks once again! :smile:

cfilby:
The ManageAuthentication warnings are now resolved - I was initially just naming the chunks since I hadn't looked into the preload yet, but I can do that with this PR as well.

cfilby:
That's fine with me - I need to work on another project in the short term but I'm hoping to rotate back to this PR in a few days.

cfilby:
Haha, well a few days was more like a few weeks. 

So I've added a ChunkName field to web.Props and added a chunkName preload test. Could you rerun the CircleCI flow - the test-ui flow failed due to the following error:

```
Build-agent version 0.1.1073-1f69f340 (2018-11-20T18:07:03+0000)
Starting container getfider/circleci:0.0.16
  image cache not found on this host, downloading getfider/circleci:0.0.16

Error response from daemon: unauthorized: authentication required
```

I'm seeing the preload assets in the head of the page - are there other ways to verify the preload is working as intended? 

Let me know if there's anything else you want me to change!

goenning:
Wow, how did I miss this? 

I'm sorry for the delay, I didn't notice your recent changes on this PR. I pulled locally and it's working as expected. If you set chrome network throttling to 'Slow 3G', you'll see chrome downloading the Home page assets before the main bundle is executed.

Just give me some time to go through the changes and review it. Thank you!

Btw, have you seen this? https://dev.to/goenning/how-we-reduced-our-initial-jscss-size-by-67-3ac0
I wrote that post based on recent work related to performance on Fider and it got to top 7 posts of last week. You're name is there 😄

goenning:
This is looking great! 

My only comments are:

1) merge from master again, there was a minor change on how we load the assets. For reference: #669 
2) there are 3 pages on `router.go` that use `handlers.Page(title, description)` instead of a custom handler function. We should add the chunkName to those too.

cfilby:
No problem! I went ahead and added a chunkName parameter to handler.Page and added chunkNames to those other pages in routes.go. 

And I actually didn't catch that, thanks for the shoutout! I'm excited to check out the full article tomorrow!

goenning:
Great PR, thank you!

Issue #744, opened 2019-03-21
Mailgun has 2 base URLs (US and EU)


Khazuyo:
Hey there!

I've - so far - successfully set up a local Fider installation and hooked it up to Mailgun. Within the sandbox everything's okay but as soon as I went into productive mode and tried sending via a real domain, this happened:

`app_1  | DEBUG [2019-03-21T14:54:00Z] [WEB] [Hrtp1bfMHsjwMJZAXwEf3owvoRsrmHfd] Email sent with response code 404.`

It turns out that Mailgun as 2 base URLs now
- https://api.eu.mailgun.net/v3/ for "EU domains"
- https://api.mailgun.net/v3/ for "US domains"
and sadly they aren't exchangable.

Using the "correct" API via curl on the command line works fine.
My Fider installation, however, can't send mails because the URL seems to be hardcoded (app/pkg/email/mailgun/mailgun.go) and the US API doesn't know about my EU domain.

I assume that I will have to recompile Fider myself if I want to fix this?

goenning:
Good point. We have hardcoded the US endpoint. Multi-region was not supported when we did the integration.

And you're also correct, you'll need to change the source and recompile if you want to use the EU domain.

If you want to contribute, I would happy to accept a PR with that option being configurable :)

I can see a new configuration `EMAIL_MAILGUN_REGION` with valid options as blank, EU or US. Then mailgun.go would use the correct domain based on configuration. A blank setting would default to US.

Khazuyo:
Well, I am already setting up a development VM on my Windows PC because I kind of want to do more than just compiling Fider on my server just to hot-swap the binary.
I'll try coming up with something and will start a PR when I think that the patch is mature enough.
Not sure how fast I'll progress though, I've never worked with Go before.

Khazuyo:
Quick heads up: I already have the code running on a server I am using for a game dev project I'm leading. If you are curious, you can look at it in a separate branch, "mailgun-region-env" which is in my fork of the project's repo.

I haven't created a PR yet because some things are missing, namely the tests, but I'll catch up with that as soon as I am able to. (:

goenning:
@Khazuyo that's an amazing work, thanks very much! Looking forward to the PR 😀

by the way, did you manage to get Fider running on Windows? Because I though that was not possible yet (ref: https://github.com/getfider/fider/issues/434).

Khazuyo:
[...]

PR #745, opened 2019-03-25
Configurable Mailgun base URL

Khazuyo:
Fix for #744

When creating domains in Mailgun, you can specify it to be hosted in either the US or the EU region. This also changes the API URL used when sending mails which was not configurable.
This PR introduces code which lets you configure the base URL used for talking to Mailgun by setting the env var EMAIL_MAILGUN_REGION to either "US" or "EU".


Khazuyo:
Here's the PR, I hope I did not forget anything especially since I'm a complete newbie to Go x)

app/pkg/email/mailgun/mailgun.go
```diff
@@ -33,6 +39,30 @@ func (s *Sender) Send(ctx email.Context, templateName string, params email.Param
...
+
+	// Default to the US domain if no region code was provided (ENV not set)
+	// or if the provided code isn't valid
+	if len(regionCode) < 1 {
+		regionCode = "US"
+	} else if len(baseURLs[regionCode]) < 1 {
+		s.logger.Warnf(
+			"Unknown Mailgun region code '@{Code}' configured - falling back to 'US'", 
+			log.Props{
+				"Code": env.Config.Email.Mailgun.Region,
```
goenning: you can reuse regionCode variable here :)
```suggestion
				"Code": regionCode,
```

app/pkg/email/mailgun/mailgun_test.go
```diff
@@ -198,3 +198,31 @@ func TestBatch_Success(t *testing.T) {
 	</body>
 </html>`)
 }
+
+func TestGetBaseURL(t *testing.T) {
+	// Memorize the original setting since we are about to change it
+	var initialRegion = env.Config.Email.Mailgun.Region
```
goenning: You must just call `RegisterT(t)` at the beginning of the test like all other tests. If you try to run this test in isolation it'll fail with a message that you forgot to call RegisterT.

Another benefit of RegisterT is that it'll automatically restore `env.Config` object, so you don't need to manually memorize and restore it :)

app/pkg/email/mailgun/mailgun.go
Khazuyo: Ah, I think I should have commented that line since it's not that obvious.
If I were to set the region code to something like "Usa" the error message would tell me that the code 'USA' is unknown which might appear strange since I used a mixed case value.
I decided to not reuse regionCode at this point to prevent any confusion.

From a functional standpoint it's mostly if not completely cosmetic though, if you still think it's better to change it I can do that, np.

app/pkg/email/mailgun/mailgun_test.go
Khazuyo: Ah, no wonder I did not notice any issues, I always ran at least all of the Mailgun tests, never just a single one. Good to know though, I'll fix it (:

app/pkg/email/mailgun/mailgun.go
goenning: That's a good point, haven't thought about that. Let's keep it as is :)

goenning:
Alright, so just remove `initialRegion` and we merge it. That's not necessary anymore after using RegisterT :)

Khazuyo:
I removed the obsolete code, I think everything should be set now?

goenning:
Merged! Thanks for the contribution!

Khazuyo:
You are most welcome! (:

Issue #769, opened 2019-08-18
Add fake SMTP server to docker-compose


renannprado:
The goal here is to allow easier local development setup.

PR #770, opened 2019-08-18
Add fake SMTP server to docker-compose

renannprado:
**Issue:**  #769
Now the docker-compose file contains also contains a fake SMTP server
to be used while developing. This should make setting up local development
environment even easier.
The contributing documentation was also updated to reflect these changes.


goenning:
nice one, thanks 👍

PR #774, opened 2019-10-15
update markdown service to avoid malicious redirects

melbarch:
**Issue:**  Secure embedded urls when converting from markdown.

When linking to other links through target=_Blank,The other page can access your window object with the window.opener property. This exposes an attack surface because the other page can potentially redirect your page to a malicious URL [more info](https://developers.google.com/web/tools/lighthouse/audits/noopener)

**Changes:** : add rel attribut to anchors


goenning:
Nice work, thanks for the PR! Don’t worry about the broken build, I’ll get sorted first and then come back to this 👍

melbarch:
We are almost there 😄

Issue #775, opened 2019-10-15
Migrate from tslint to eslint


goenning:
might be a good first issue for people doing the Hacktoberfest 🙂

dvfleet413:
@goenning I'd be willing to work on this if it's still open

goenning:
Yes, it’s still open. By the way, we want to continue using prettier as well, so the plan is to migrate from tslint+prettier to eslint+prettier

dvfleet413:
Great, thanks for clarifying. I’ll work on it tonight, should have a PR in the next few days.

PR #777, opened 2019-10-16
Migrate to ESLint

dvfleet413:
In reference to Issue #775 

Uninstalled and removed TSLint
Installed ESLint, eslint-plugin-prettier, and eslint-config-prettier

This was mostly done from the command line, with a few changes to the eslint configuration file to (hopefully) prevent issues with ESLint and Prettier working together.  Let me know if any changes need to be made.


package.json
```diff
@@ -46,7 +50,7 @@
     "tslint-react": "3.6.0",
```
goenning: `tslint-react` and any other tslint package needs to be removed

dvfleet413:
Thanks for the feedback! And for being patient with a newbie :)  I was able to fix most of the issues, but wasn't positive about updating the `Lint` function in magefile.go.  Here's what I have:
```
func Lint() error {
      return sh.Run("npx", "eslint", "-c", ".eslintrc.json", "'public/**/*.{ts,tsx}'", "'tests/**/*.{ts,tsx}'")
}
```
Should that work? Once I get that function fixed I should be good :)

goenning:
I’m not an expert on eslint either, but it looks ok to me. Have you tried to run it locally? You’ll need to install Go 1.12 and Mage (see CONTRIUITING.md). You can the run it with: `mage lint`

PR #779, opened 2019-10-18
Migrate to eslint

dvfleet413:
This one seems to be building okay in my IDE, but I'm having a hard time setting up the local environment to actually run the program.  The packages that were updated have been reverted back to where they were, and the `Lint` function in `magefile.go` seems to be working.


goenning:
Can you check the Build? It's failing on the lint step due to invalid json on .eslintrc.json

PR #782, opened 2019-10-22
tooling: Use NodeJS 12 in NVM config

mdemierre:
NodeJS 12 is documented in the `CONTRIBUTING` guide, so this updates NVM to use `v12.*`


goenning:
Great, thanks for that 👍

PR #787, opened 2019-10-25
Replace deprecated StatelessComponent to FunctionComponent

chagriali:
This PR renames the deprecated React StatelessComponent  to FunctionComponent.


goenning:
Thank you :)

Issue #837, opened 2020-01-07
Name tooltip on avatar image (via title)


esgn:
Think there's something missing in https://github.com/getfider/fider/blob/master/public/components/common/Avatar.tsx 

I  added "name: string;" in AvatarProps>user and set title to {props.user.name} in the img tag. Seems to work ok in order to get the username via simple tooltip on mouse over.

`interface AvatarProps {
  user: {
    role?: UserRole;
    avatarURL: string;
    name: string;
  };
  size?: "small" | "normal" | "large";
}`

`return <img className={className} title={props.user.name} src=.../>;`

goenning:
hey @esgn, yep, apparently we missed that property. Want to send a PR?

esgn:
Ok. I will very soon. Thx

PR #842, opened 2020-02-04
Patch on Avatar.tsx to activate name tooltip via title attribute

esgn:
**Issue:**  https://github.com/getfider/fider/issues/837

User name added to props and title attribute updated to display basic name tooltip on avatar pic.


goenning:
Thanks for the PR! 😀

PR #847, opened 2020-03-19
app/models: extract side effects

carnott-snap:
**Issue:**  #848

Currently importing app/models triggers app/pkg/jwt, which causes the
init in app/pkg/env to invoke. While this may be a foregone conclusion
in some contexts, data models should be pure.

This change extracts the jwt model into a new package app/models/jwt,
while forwarding all refrences, except app/models.CreateTenant, through
the old app/pkg/jwt, to reduce noise and because one package is a
better interface than two.

### DIFF
--- a/app/models/identity.go
+++ b/app/models/identity.go
@@ -4,7 +4,7 @@ import (
 	"time"
 
 	"github.com/getfider/fider/app/models/enum"
-	"github.com/getfider/fider/app/pkg/jwt"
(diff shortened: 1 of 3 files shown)


arnottcr:
any reason for the closure?

goenning:
No reason. I think it was closed when I renamed the branch from master to main. I can't reopen because:

Screenshot 2021-03-30 at 19 12 41

PR #862, opened 2020-05-04
Define deployment

maxvu:
Add Helm chart, deployment info.


helm/Chart.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+appVersion: "1.0"
+description: A Helm chart for Krang
```
jesse-welch, line 3: Gotta make sure we find and fix all references from other projects

PR #874, opened 2020-08-13
tooling: build multiarch docker image

abeMedia:
Currently the docker images only work on `amd64` architecture. This PR adds support for `arm64` and `386`.


goenning:
Thanks for the PR and sorry for being slow to respond.

I’m new to buildx so I had to do some research. Have you tried this action here before? https://github.com/docker/build-push-action

I’d prefer using the official Docker actions if possible. Apparently we could even replace that azure docker login as well (which I honestly don’t know I use that...)

abeMedia:
Hi @goenning, thanks for the response.
That action only came out shortly before I did this PR so I hadn't heard of it yet at the time.
Happy to amend this PR to use that though.

Dockerfile
```diff
@@ -32,7 +34,6 @@ FROM debian:buster-slim
 RUN apt-get update
 RUN apt-get install -y ca-certificates
 
-RUN mkdir /app
```
abeMedia, line 35: Removed these as `WORKDIR` creates the folder if it doesn't exist: https://docs.docker.com/engine/reference/builder/#workdir

goenning: thanks, I didn't know that :+1:

goenning:
@abeMedia thanks for the update. the reason why I went with a separate `workflow_run` was to publish an image for each pull requests as well, so I'd like to keep that.

- push to main getfider/fider:main and getfider/fider:SHA_{sha7}
- pull request: getfider/fider:PR_{number}

abeMedia:
@goenning still not working as `v8go` doesn't support these architectures. Currently working on an upstream PR for that. 
Is that a new dependency btw? Because I've been using the ARM version in production for months.

As for the PR images that's being done via [`crazy-max/ghaction-docker-meta`](https://github.com/crazy-max/ghaction-docker-meta#semver) which I stumbled across on the [`docker/build-push-action` docs](https://github.com/docker/build-push-action/blob/master/docs/advanced/tags-labels.md). The naming is slightly different that your current one though (`pr-$N` instead of `PR_$N`).

goenning:
1. Yes, v8go is a recent addition.
2. That won’t work because the workflow doesn’t have access to secrets, which is why I used workflow_run as recommended here https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

abeMedia:
OK will change that once the upstream issue is fixed and this is working.

urunimi:
@abeMedia 
It looks like [this PR](https://github.com/rogchap/v8go/pull/202) is merged. Do you think this will resolve the `v8go` issue?
I also need fider for arm64.

goenning:
I think we’ll need to wait for https://github.com/rogchap/v8go/pull/223 as well

abeMedia:
I've been using this branch on ARM for months now. All worked fine.

urunimi:
> I've been using this branch on ARM for months now. All worked fine.

Oh, thanks! I'll try to build an image from this branch. @abeMedia 👍

PR #879, opened 2020-09-05
fixed a bug where users do not receive notifications for comments

### LINKED ISSUE #878: Subscribed users don't receive any notifications after added comments

HosseinYousefi:
**Issue:** closes  #878

added just a single line that we had for the status change but not for added comments.


goenning:
@HosseinYousefi thanks for the PR, that was the initial implementation, but I had to remove because a lot of people were complaining about unsolicited email updates.

I’m thinking that we should have a small checkbox when adding a comment so the user can confirm if they want to subscribe or not, which of course would only be visible if they have not yet subscribed.

PR #895, opened 2021-01-03
security: bump npm and resolve host attack

goenning:


tziemek:
@goenning How can we use Fider behind a reverse proxy now? It is always returning a 404 if the request URL and the hostname do not match.

tziemek:
Never mind, I think our reverse proxy isn't configured correctly.

PR #916, opened 2021-03-14
Update package.json

bwaj:
**Issue:**


goenning:
@bwaj i think I missed this change, but should it be like this maybe:

“NODE_ENV=production npx webpack-cli”

Issue #925, opened 2021-03-22
Connecting to SMTP server using certificate from private CA


gs11:
I've set up Fider to run in an internal Kubernetes cluster where it will use the company internal SMTP server for emails.
This server uses a certificate issued by a private company CA which causes certificate validation to fail.
I could not find any way to disable TLS either so a PR is incoming.

PR #926, opened 2021-03-22
fix: Allow SMTP without StartTLS

gs11:
Fixes https://github.com/getfider/fider/issues/925

Disclaimer, I've not written a single line of Go so I'm very open to improvement suggestions (if any) 😅 
Tested OK in local setup.


goenning:
Code looks good to me! I’d just suggest naming it `..._ENABLE_STARTTLS` and make it true by default. It’s often easier to read code with positive variables instead of negating a negative var

gs11:
Changed as suggested @goenning

goenning:
@gs11 thanks for your contribution. Did you have any issues when setting up Fider locally? I’m trying to simplify where possible, so any feedback is appreciated.

gs11:
Thanks!
No issues really but minor hurdles:
* Required `pg_trgm` extension disallowed us to use our internal shared DB cluster
* Undocumented environment variables (such as `SIGNUP_DISABLED`)

Other than that, very smooth experience with great docs!

Issue #949, opened 2021-04-30
Translate to japanese


PeterTakahashi:
Do not use this form for feature requests, please vote or post it on https://feedback.fider.io/

goenning:
I'm going to close this in favour of #575

PR #950, opened 2021-04-30
[WIP]React Translate to Japanese

PeterTakahashi:
**Issue:** 
https://github.com/getfider/fider/issues/949

# screen shot

## SignUp Page
スクリーンショット 2021-04-30 14 40 46

## HomePage
Japanese
スクリーンショット 2021-04-30 14 49 21

English
スクリーンショット 2021-04-30 14 50 02

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12388 +1,61 @@
 {
   "name": "fider",
   "version": "0.0.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
   "requires": true,
-  "packages": {
-    "": {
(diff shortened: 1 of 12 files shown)


goenning:
That's very promising, thanks for sharing it. I'll have a better look and try locally when I get more time. But how do you see this working on the backend and emails? Maybe we could read the same jsons files used by the frontend?

PeterTakahashi:
> Maybe we could read the same jsons files used by the frontend?

Yes, it is.
I guess I should make sure that Go and React see the same json file.

Issue #880, opened 2020-09-10
URLs of custom OAuth providers transformed to lowercase automatically


tobiashinz:
I tried to connect my Fider instance to Keycloak.

My Keycloak realm has a capital letter in its name. Fider transforms the Authorize URL, Token URL and Profile API URL to all lowercase automatically. 

This way it is a realm-name mismatch with Keycloak and the authentication fails.

I was able to work around it by changing the mentioned URLs in the database.

spadge4711:
Same for me... Unable to use it with oauth2 though

goenning:
Should be an easy fix, I didn’t know URLs are case sensitive. Anyone interested in submitting a PR?

indyteo:
I can do it, it doesn't seem to be much complicated: #973

goenning:
Thanks for the contribution @indyteo 👍

PR #973, opened 2021-07-01
fix: removed auto-lower on URLs (OAuth)

indyteo:
**Issue:** #880 

Removed the automatic lowercase on OAuth `AuthorizeURL`, `TokenURL` and `ProfileURL`, as URLs are case-sensitive


goenning:
Thanks 🎉

Issue #804, opened 2019-11-04
Allow admins to select which sign in providers to enable


goenning:
We should have an option for admins to disable the sign in options that are enabled by default. A possible implementation is to create a new `tenant_settings` table similar to `user_settings`, which would accomodate generic settings such as:

| Name  | Value |
| ------------- | ------------- |
| sign_in_email_enabled  | true/false  |
| sign_in_oauth_google_enabled | true/false  |
| sign_in_oauth_facebook_enabled | true/false  |
| sign_in_oauth_github_enabled | true/false  |

gcase555:
Can I volunteer for this one? I mean someone else could technically pick it up but I am interested in this feature and would like to avoid duplicate effort if its already being worked on.

TimNilimaa:
> @goenning Can I volunteer for this one? I mean someone else could technically pick it up but I am interested in this feature and would like to avoid duplicate effort if its already being worked on.

Also interested in this, could we do it together?

automoto:
We'd have to figure out some kind of collaborative workflow or how to break it up.

I can do some frontend and backend so I can do either.

chrnin:
Nice idea, I have a similar need. For instance, here a use a private identity provider and it would be very useful to have such an automatic sign-in feature for my in-house provider. The way it is described with fixed table fields for each provider seems a little reductive since in my case I couldn't mention it for a custom self-made identity provider. 
i'm also interested into participating on this one… My idea would rather be to set a boolean parameters on identity providers configuration to enable automatic sign-in when someone logs in using it. 
Would that fit your needs ?

chrnin:
Oops, I think I misunderstood… My need is rather getting automatic account creation even if the fider instance is private…

mattwoberts:
This is pretty stale now, there are a good few options for enabling / disabling oauth providers in Fider, so closing this.

PR #976, opened 2021-07-05
feature: added control over email authentication for admins

indyteo:
**Issue:** #804 (partially)

To get detailed information, please read the discussion https://github.com/getfider/fider/discussions/969#discussioncomment-956072 (first part of message, **Control over email authentication**)


indyteo:
I can't manage to push a simple feature without doing several merge commits, it's annoying I don't know why there are so much of them, I'm sorry :frowning:

migrations/202106301200_added_allowing_email_auth.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE tenants ADD COLUMN is_allowing_email_auth BOOLEAN;
```
goenning, line 1: I'd suggest `is_email_auth_allowed`, it's easier to follow that pattern in future if we need: `is_{something}_allowed` or `is_{something}_enabled` etc

migrations/202106301200_added_allowing_email_auth.sql
```diff
@@ -0,0 +1,3 @@
...
+UPDATE tenants SET is_allowing_email_auth = TRUE;
```
goenning, line 2: There was a migration file created on the previous PR with a date before this one, so we need this one to be moved to `20210720...`

public/components/common/SignInControl.tsx
```diff
@@ -77,6 +78,23 @@ export const SignInControl: React.FunctionComponent<SignInControlProps> = (props
...
+              <Trans id="signin.message.onlyadmins">Currently only allowed to sign in to an administrator account</Trans>
+            </p>
+          )}
+        </div>
+      ) : (
+        <div>
+          <p className="text-muted">
+            <Trans id="signin.message.emaildisabled">
+              Email authentication has been disabled by an administrator. If you have an administrator account and need to bypass this restriction, please{" "}
+              <span className="text-bold clickable" onClick={() => setUseEmail(true)}>
```
goenning, line 92: I think we could turn this into a `<a href="#" className="text-link"` ... instead of a clickable span. I'm slowing removing all occurrences of that because of accessibility issues.

public/pages/SignIn/SignIn.page.tsx
```diff
@@ -44,7 +44,7 @@ export default class SignInPage extends React.Component<any, any> {
         </div>
         <div className="text-center w-max-4xl mx-auto mb-4">{Fider.session.tenant.isPrivate ? <Private /> : <Locked />}</div>
 
-        <SignInControl onEmailSent={this.onEmailSent} useEmail={true} redirectTo={Fider.settings.baseURL} />
+        <SignInControl onEmailSent={this.onEmailSent} useEmail={Fider.session.tenant.isAllowingEmailAuth} redirectTo={Fider.settings.baseURL} />
```
goenning, line 47: I'm not sure we can reuse the useEmail to be honest. If you look at signup.page.tsx, you'll notice that it has some problems there.

Screenshot 2021-07-20 at 21 07 40

public/pages/Administration/pages/ManageAuthentication.page.tsx
```diff
@@ -55,69 +60,125 @@ export default class ManageAuthenticationPage extends AdminBasePage<ManageAuthen
...
-                <HStack className="h-6">
-                  <OAuthProviderLogo option={o} />
-                  <strong>{o.displayName}</strong>
+      <VStack spacing={8}>
+        <div>
+          <h2 className="text-display">General Authentication</h2>
+          <Form error={this.state.error}>
+            {this.state.isAllowingEmailAuth &&
+              (this.state.canDisableEmailAuth ? (
+                <p className="text-green-700">You can disable email auth because you have at least one authentication provider enabled</p>
```
goenning, line 112: I think we simplify the messaging. This green text being displayed at all times is unnecessary in my option. I'm thinking of:

1. remove this green/red sub-message
2. as a helper text of the checkbox, if the input is disabled, explain why. If the checkbox is enabled, explain what happens when that is toggled off.

The note can be displayed at all times.

app/actions/tenant.go
```diff
@@ -202,3 +204,29 @@ func (action *UpdateTenantPrivacy) IsAuthorized(ctx context.Context, user *entit
...
+func (action *UpdateTenantAllowingEmailAuth) Validate(ctx context.Context, user *entity.User) *validate.Result {
+	result := validate.Success()
+
+	activeProviders := &query.ListActiveOAuthProviders{}
+	if err := bus.Dispatch(ctx, activeProviders); err != nil {
+		return validate.Failed("Cannot retrieve OAuth providers")
+	}
+
+	if len(activeProviders.Result) == 0 {
+		result.AddFieldFailure("isAllowingEmailAuth", "You cannot disable email authentication without any other provider enabled.")
```
goenning, line 228: what about `You need to configure another authentication provider before disabling email authentication.`

app/actions/oauth.go
```diff
@@ -47,6 +48,18 @@ func (action *CreateEditOAuthConfig) IsAuthorized(ctx context.Context, user *ent
...
 
+	if action.Status == enum.OAuthConfigDisabled {
+		tenant := ctx.Value(app.TenantCtxKey).(*entity.Tenant)
+		activeProviders := &query.ListActiveOAuthProviders{}
+		if err := bus.Dispatch(ctx, activeProviders); err != nil {
+			return validate.Failed("Cannot retrieve OAuth providers")
+		}
+
+		if !tenant.IsAllowingEmailAuth && len(activeProviders.Result) == 1 {
+			result.AddFieldFailure("status", "You cannot disable this provider with neither email auth nor any other provider enabled.")
```
goenning, line 59: is there a situation where the UI would allow this to happen?

goenning (changes requested):

goenning:
no problem with the multiple merge commits, I always squash every PR

migrations/202106301200_added_allowing_email_auth.sql
indyteo: Yes, I saw the `..._add_locale_field` that came right after this one :sweat_smile: 
Renamed to `202107211126_...`

migrations/202106301200_added_allowing_email_auth.sql
indyteo: I agree, that name wasn't really pretty, it'll be better like this :smile: 
Renamed database column to `is_email_auth_allowed` in table `tenant`
Renamed field to `IsEmailAuthAllowed` in Go structures
Renamed field to `isEmailAuthAllowed` in JSON models
Renamed several functions using a related name in both back-end and front-end

public/components/common/SignInControl.tsx
indyteo: No problem, I see, I changed it!

public/pages/SignIn/SignIn.page.tsx
indyteo: Oh damn I didn't noticed this one :frowning: 
I'll introduce a new state, it's not a big deal
Rolled back uses of `useEmail` and added `showEmailForm` state

app/actions/tenant.go
indyteo: Yes of course, this is far better that my original sentence, my English level produce sometime really weird wordings :sweat_smile: 
Changed message

app/actions/oauth.go
indyteo: In theory, no. But users may force the checkbox, or maybe a bad synchronization due to a poor connection from the client would cause him to disable email authentication switch and then rush into the OAuth provider edit form (I don't know if it's really possible but I think so, with a very slow connection I think), so I prefer to add this check as we never know what could possibly happen.
Or even worse: Another admin disable the email authentication while the other is on the OAuth provider edit form, and then his UI become out of sync with the server.

public/pages/Administration/pages/ManageAuthentication.page.tsx
indyteo: Ok I agree the messaging wasn't optimal, is that better like this?
Please let me know if you have any comment about it now :blush:

public/components/common/SignInControl.tsx
```diff
@@ -15,9 +15,15 @@ interface SignInControlProps {
 
 export const SignInControl: React.FunctionComponent<SignInControlProps> = (props) => {
   const fider = useFider()
+  const [showEmailForm, setShowEmailForm] = useState(fider.session.tenant.isEmailAuthAllowed)
```
goenning, line 18:
```suggestion
  const [showEmailForm, setShowEmailForm] = useState(fider.session.tenant ? fider.session.tenant.isEmailAuthAllowed : true)
```

this control is used for signup, so we'll need this.

goenning (changes requested):
Looks good to me :+1:

I just suggested one fix and also now that SignInControl doesn't include it when useEmail is false, we need to add a Divider into the SignUp.page.tsx.

```
            <SignInControl useEmail={false} />
            <Divider />
            <Form error={this.state.error}>
              <Input field="name" maxLength={100} onChange={this.setName} placeholder="Name" />
              <Input field="email" maxLength={200} onChange={this.setEmail} placeholder="Email" />
            </Form>
```

Thanks again for updating it, we'll merge it this week and get it live 🚀

JohnnyMorganz:
Hey, thanks for sending a PR for this - it will be great help for us in our fider instance.
I just wanted to point out a few things which you may or may not have already solved - if you have already solved this, and I missed it, I do apologise!

1) Some of the API routes don't seem to be protected against a check to see if email sign in is disabled. I think this is mainly the sign up route. This is an issue for us since a malicious user may use these routes to sign in which we do not want to happen (everything we do is through SSO). It also sends out verification emails, which we don't want to happen.

2) I'm not sure if this is *completely* correct, but should changing `email` on the MySettings page be disabled if email sign up is disabled? These emails (for us, atleast), are linked to our SSO flow, and we don't want users to change them in the fider instance.

3) Similar to 2., but we want to enforce that the name used in fider is always the same as the one in our OAuth SSO - so for us, changing names in the mysettings page shouldn't be allowed. This may be a separate issue though - which can be solved in a different PR. (A separate issue we *do* have here is that if the username is changed in SSO it isn't updated in fider, but that's unrelated here)

Again, sorry if you've already covered some of these points, I only had a quick glance so I may have missed it! And thanks again for implementing this, its very useful for us.

indyteo:
Hey,
Thanks for having a look at the PR :blush: 

Concerning the first question, I think I covered the only entry-point available because this is the only code generating a sign in key. Of course the key verification end-point don't verify the setting, but isn't unnecessary since users cannot generate a key. And keys are valid only 30 minutes, so we can't keep a key and use it later (but this setting is not intended to be switched every 30 minutes, it's rather a one-time for cloud-based hosting, like the locale).

I think you misunderstood something with your next question. The goal is not to remove every email from Fider. The OAuth provider should give an email per user, so why wouldn't they be able to opt-in for email notifications? If this is related to your particular company that don't want to send emails, yes this is another question. But disabling email authentication shouldn't prevent users from receiving email notifications according to me.

And finally for the unmodifiable username that could be an interesting feature combined with this one your right... Maybe an extra setting here, even if I don't feel comfortable with the idea of preventing a user from editing his name, but never mind.

Don't apologize for submitting feedback, it's important to make things eve better! :smile:

JohnnyMorganz:
Thank you for your response!

> I think you misunderstood something with your next question. The goal is not to remove every email from Fider. The OAuth provider should give an email per user, so why wouldn't they be able to opt-in for email notifications? If this is related to your particular company that don't want to send emails, yes this is another question. But disabling email authentication shouldn't prevent users from receiving email notifications according to me.

I don't mean disabling email notifications here, I mean disabling **changing the email** once you have signed up. Currently, when signing in using OAuth, you can modify the email (and username) you signed up with in the my settings page.
This can become an issue if someone changes their email - as it becomes out of sync with our SSO

> Maybe an extra setting here, even if I don't feel comfortable with the idea of preventing a user from editing his name, but never mind.

The idea isn't actually to *prevent* a user changing their name, but that the name is always the same as their SSO name. For us, a user can still change their name via their SSO account, but we don't want them to have a different name on the fider instance and everywhere else!

indyteo:
> I don't mean disabling email notifications here, I mean disabling changing the email address once you have signed up.

Oh so it was me who misunderstood this one :sweat_smile: 
So I think we could envisage a better SSO management with the following elements:
* A setting to enable it when email authentication is disabled
* The ability to register a "SSO settings URL" such as `https://example.org/sso/settings` or even better, with the reference ID as a wildcard `https://example.org/sso/{id}/settings`
* When enabled, the name and email settings become unmodifiable and the URL is shown with a notice as you said
* An API end-point to edit username and email that can be called by the SSO system when user updates its settings

It seems interesting, but as you stated before, it's another issue, even if it's related.

JohnnyMorganz:
Definitely, better to get this PR sorted and then these things can be solved at a later date!

> The ability to register a "SSO settings URL" such as `https://example.org/sso/settings` or even better, with the reference ID as a wildcard `https://example.org/sso/{id}/settings`

There is already a `Profile API url` setting when setting up OAuth, which provides the username and email information. I talk about the sync issue in #986.

I think the design can be simpler, as you mention it can be just a toggle setting in the admin panel such as "Manage username and email through OAuth", and then:
- The username and email can be automatically retrieved from the endpoint mentioned above, so I don't think a "SSO settings URL" is necessary
- The username and email are automatically resynced when a user logs in again, as mentioned in #986. This removes the need of creating a new API endpoint for the SSO system to call, fider will just automatically update itself once a user signs in again.

goenning:
In my view there's a difference between what we support on OAuth and what companies want with SSO.

OAuth is great for public usage, maybe your audience is Gaming, so you could add twitch and discord to reduce the signup barrier and get more feedback. But users should be able to change their name/email and not have this permanent link to the source platform. In this model OAuth is just serving as a "quick" authentication. 

As for SSO, it's usually used to restrict and facilitate access from company employees to a website/platform. In this scenario you're right that ideally names/emails are managed centrally so we should not allow users to change them, but rather sync with the source.

Fider has a great support for OAuth, but a true SSO is missing.

JohnnyMorganz:
Our audience actually is gaming 🙂, our "SSO" (maybe that's the wrong word here) is used to tie player's game accounts to the feedback forum, so their fider username etc. == their player username.
It helps us link the two together, and also for moderation purposes.

I think the feature above reasonably covers the issue currently, however, I completely understand if that's not a focus for the project.

PR #893, opened 2020-12-15
feat: add webhook

acamilleri:
**Issues:**  
- https://github.com/getfider/fider/issues/892
- related https://github.com/getfider/fider/issues/882

Issue #892, opened 2020-12-15
Add webhook


acamilleri:
Feature request: 
- https://feedback.fider.io/posts/36/add-slack-integration
- https://feedback.fider.io/posts/185/webhook-updates-on-events (Partial implementation)
--- 

I added a Webhook service, to received notifications (new_post, new_comment, status_change and delete_post).
It's not User facing, this first version it's only be configured with the `.env` file.

I create an Notifier interface, to easily add new providers (Microsoft Teams, Mastodon, ...).

urinal-cake:
Is it possible to sponsor the webhook implementation? My business could ideally use this as soon as possible for reporting information on new suggestions, comments, status changes, deleted posts _and_ if a user upvoted/downvoted ideas. We use Discord, and would organize the information in conjunction with Stackstorm.

FOSSforlife:
If this helps anyone, I wrote up a [Node.js script](https://github.com/FOSSforlife/fider-webhook) that fetches updates from the Fider API every few minutes and posts them to a webhook. Until webhook functionality is official, this works for me in the meantime.

PR #980, opened 2021-07-20
feature: add webhooks customizable from admin panel

indyteo:
**Issue:** #892 / #893 / [Webhook updates on events](https://feedback.fider.io/posts/185/webhook-updates-on-events) / [Add Slack Integration](https://feedback.fider.io/posts/36/add-slack-integration)

Please check this discussion (https://github.com/getfider/fider/discussions/969#discussioncomment-1013585) to see every details.

Note that I based the general idea on what @acamilleri did in his PR and extended it from the environment variable to the admin control panel.


goenning:
I started reviewing this today, didn't went through code yet, but I love how flexible the templates are and the feature is working end to end. Great work!

My suggestions below, I would like to hear your opinion if you don't agree with some of these :)

1. HTTP Method could be a dropdown, GET/POST/PUT/DELETE/PATCH should be enough?!
2. Header Name/Value might be better in the same row
3. I was able to save a webhook with invalid content: eg: `{{ boom .post_title }}`. Is it possible to validate before saving?
4. Some tooltips have a "hand" icon but no click action. We could use the "?" or pointer cursor instead, only use the hand if there's a click action.
5. Preview looks great! It's super useful!

goenning:
btw, I had to change the migration filename to make it work

indyteo:
Thanks for starting the review!

About your questions:
1. I first planed to do this, but those methods are just the standardized ones, and a server could implement custom values: [See RFC-7231](https://datatracker.ietf.org/doc/html/rfc7231#section-4). Maybe a drop down with common values and a special "Other" value that show an additional text input field allowing a custom value.
2. What do you mean by "in the same row"? Isn't it the case now? Could you please join a screenshot explaining your expectation about this? Thanks
3. Hmm, it seems possible. We would need to parse the template in the validation, an check if an error occurred. We can consider this option.
4. I totally agree, but this is the tooltip style that define the hand cursor. I don't know why it was defined like that but if you think it was a mistake we can fix it.
5. I don't think a huge modal is a problem, in fact the length of the content won't change as we have the same things to display. But the modal allow users to scroll the help separately from the rest of the page. I think we're in fact using the best visualization method.
6. Yes, I'm pretty proud of it 😊

goenning:
> 1. I first planed to do this, but those methods are just the standardized ones, and a server could implement custom values: [See RFC-7231](https://datatracker.ietf.org/doc/html/rfc7231#section-4). Maybe a drop down with common values and a special "Other" value that show an additional text input field allowing a custom value.

Fair point, but having a dropdown and an input would just look more complicated, so let's keep it simple.

> 2. What do you mean by "in the same row"? Isn't it the case now? Could you please join a screenshot explaining your expectation about this? Thanks

image

I think header and value could be side by side, possibly a small header input and a larger value input.

> 3. Hmm, it seems possible. We would need to parse the template in the validation, an check if an error occurred. We can consider this option.

It's kind of like running the preview as a validation for it.

> 4. I totally agree, but this is the tooltip style that define the hand cursor. I don't know why it was defined like that but if you think it was a mistake we can fix it.

Ok, I see now. That's a mistake, all the existing tooltips are not clickable, so the "pointer" cursor doesn't seem like the correct one.

> 5. I don't think a huge modal is a problem, in fact the length of the content won't change as we have the same things to display. But the modal allow users to scroll the help separately from the rest of the page. I think we're in fact using the best visualization method.

Same as item 1, let's go with a simple solution and improve later.

indyteo:
It's curious, I already placed the header name and value fields on the same row
The two columns
It surely is due to a [bad browser compatibility](https://caniuse.com/css-grid), on which browser did you get this render?

I meant to force the check of the template compile to validate the configuration (i.e. not being able anymore to save an invalid template). Do you think it's a good point to allow users to save invalid configuration, maybe because it's a draft?

I'll fix it so, only the clickable elements will get the pointer cursor :smile:

indyteo:
> I think it was Safari. Maybe need to use the HStack instead of Grids?

I first had some issues with HStacks when I tried them, it was why I switched to grids, but I'll try to find a solution with HStacks

> We could validate when saving it as active, but allow invalid template if saved as inactive, what you think?

Excellent idea, it seems to be a good choice, and very simple to do. I like that

> I'll review the code this in and come back with my feedback in case you want to wait and batch changes

I'll do those two small features right now and will adapt the rest later when you'll have reviewed it, no problem!

app/services/webhook/props.go
```diff
@@ -0,0 +1,161 @@
...
+		keyPrefix + "_invitation":      tenant.Invitation,
+		keyPrefix + "_welcome_message": tenant.WelcomeMessage,
+		keyPrefix + "_status":          TenantStatus(tenant.Status),
+		keyPrefix + "_locale":          tenant.Locale,
+		keyPrefix + "_url":             baseURL,
+		keyPrefix + "_logo":            logoURL,
+	})
+}
+
+func TenantStatus(status int) string {
```
goenning, line 148: would be better to implement MarshalText/UnmarshalText/String on the enum and use `tenant.Status.String()`

all enums should have this now, but it's possible that I missed some as I added this late

app/services/webhook/webhook.go
```diff
@@ -0,0 +1,173 @@
...
+}
+
+func (s Service) Init() {
+	bus.AddHandler(triggerWebhook)
+	bus.AddHandler(triggerWebhooksByType)
+	bus.AddHandler(previewWebhook)
+	bus.AddHandler(getWebhookProps)
+}
+
+func triggerWebhook(ctx context.Context, c *cmd.TriggerWebhook) error {
```
goenning, line 42: It seems like this is only used for testing, maybe we rename it to "testWebhook" or something similar? looking at triggerWebhook it gave me the impression this was the real trigger.

by using a different name here, maybe the one belo could then be just "cmd.TriggerWebhooks"

app/services/webhook/webhook.go
```diff
@@ -0,0 +1,173 @@
...
+	result.Error = error
+
+	log.Warnf(ctx, "@{Message} (ID: @{ID:yellow}, Name: @{Name:blue}): @{Error:red}", dto.Props{
+		"Message": message,
+		"ID":      result.Webhook.ID,
+		"Name":    result.Webhook.Name,
+		"Error":   error,
+	})
+
+	webhooks := &query.MarkWebhookAsFailed{ID: result.Webhook.ID}
```
goenning, line 161: after a webhook is marked as Failed we never process it again? unless someone re-enable it?

app/services/webhook/props.go
```diff
@@ -0,0 +1,161 @@
...
+		props["post_response_text"] = "The reason _why_ this post was deleted."
+	}
+	return props
+}
+
+func DescribeUser(props dto.Props, user *entity.User, keyPrefix string) {
+	if user == nil {
+		return
+	}
+	role, _ := user.Role.MarshalText()
```
goenning, line 74: most enums should have a String() function (which comes from the Stringer interface), have a look at the AvatarType enum, doing that means we can inline this on the function below

app/services/webhook/props.go
```diff
@@ -0,0 +1,161 @@
...
+	props.Append(dto.Props{
+		keyPrefix + "_id":     user.ID,
+		keyPrefix + "_name":   user.Name,
+		keyPrefix + "_email":  user.Email,
+		keyPrefix + "_role":   string(role),
+		keyPrefix + "_avatar": user.AvatarURL,
+	})
+}
+
+func DescribePost(props dto.Props, post *entity.Post, keyPrefix, baseURL string, complete, author bool) {
```
goenning, line 84: instead of baseURL here, why not get the full url from the task? we'd avoid building the url in multiple places

goenning: I'm finding a bit confusing the use of complete and author params. What do they mean?

migrations/202107061359_create_webhooks.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE IF NOT EXISTS webhooks (
+  id SERIAL PRIMARY KEY,
+  name VARCHAR(60) NOT NULL,
```
goenning, line 3: name is 60 here and 100 on validation. I think 60 is enough, so just align the number there please

migrations/202107061359_create_webhooks.sql
```diff
@@ -0,0 +1,12 @@
...
+  type SMALLINT NOT NULL,
+  status SMALLINT NOT NULL,
+  url VARCHAR(256) NOT NULL,
```
goenning, line 6: same for URL, I'd also suggest using 300 to align with the other tables and update the action validation as well

migrations/202107061359_create_webhooks.sql
```diff
@@ -0,0 +1,12 @@
...
+  content TEXT NULL,
+  http_method VARCHAR(16) NOT NULL,
```
goenning, line 8: same

migrations/202107061359_create_webhooks.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE IF NOT EXISTS webhooks (
```
goenning, line 1: and this file will have to increment its timestamp to be executed

goenning (commented):
oh wow, this is a big PR :D

I went through all the backend files, will review frontend later this week.

goenning:
Overall looks really good! Happy to see this being registered as a service :) 

I have mixed feelings about the DescribeTenant / DescribePost / DescribeUser but I can't think of a cleaner solution at the moment. The only thing that comes to mind would be a webhook specific props that you'd do:

```
p := webhook.Props{}
p.SetUser(user, ..)
p.SetPost(user, ..)
p.Set("something", ..)
```

Not sure what benefits we'd get, but at least we wouldn't have the public generics Describe* that are meant to be used only on this scenario.

public/assets/styles/utility/display.scss
```diff
@@ -103,6 +103,10 @@
   justify-content: space-evenly;
 }
 
+.justify-full > * {
+  flex: 1 1 0px; // px unit is required to support IE
```
goenning, line 107: we don't support IE :)

public/components/common/Loader.tsx
```diff
@@ -3,12 +3,22 @@ import "./Loader.scss"
...
   const [show, setShow] = useState(false)
 
   useTimeout(() => {
     setShow(true)
   }, 500)
 
-  return show ? <div className="c-loader" /> : null
+  return show ? (
+    <div className={props.className}>
+      <div className="c-loader" />
```
goenning, line 20: the `.c-{name} is often the top level element returned by the component. Can we add the className to the existing  div?

public/pages/Administration/components/webhook/WebhookForm.tsx
```diff
@@ -0,0 +1,286 @@
...
+        placeholder="Header"
+        disabled={props.onRemove !== undefined}
+        suffix={duplicate ? "Duplicate" : undefined}
+      />
+      <Input field={`value-${header}`} value={value} onChange={setValue} placeholder="Value" disabled={!editing} suffix={suffix} />
+    </HStack>
+  )
+}
+
+export class WebhookForm extends React.Component<WebhookFormProps, WebhookFormState> {
```
goenning, line 91: I'm slowly migration all older class components into functions+hook. Do you mind converting this one so we don't add new tech debt?

public/pages/Administration/components/webhook/WebhookForm.tsx
```diff
@@ -0,0 +1,286 @@
...
+          <HStack>
+            <Button variant="primary" onClick={this.handleSave}>
+              Save
+            </Button>
+            <Button onClick={this.handleCancel} variant="tertiary">
+              Cancel
+            </Button>
+          </HStack>
+        </Form>
+        <WebhookTemplateInfo type={this.state.type} isModalOpen={this.state.isModalOpen} onModalClose={this.hideModal} />
```
goenning, line 282: rename component to WebhookTemplateInfoModal to follow the convention

public/pages/Administration/components/webhook/WebhookProperties.scss
```diff
@@ -0,0 +1,32 @@
...
+    padding: sizing(3);
+  }
+
+  td {
+    border-top: 1px solid get("colors.gray.100");
+    padding: sizing(2);
+  }
+
+  &__prop {
+    font-family: $font-code;
```
goenning, line 16: there's a `get("font.family.code")`

public/pages/Administration/pages/ManageWebhooks.page.tsx
```diff
@@ -0,0 +1,180 @@
...
+      <h2 className="text-display">{props.title}</h2>
+      <p className="text-muted">These webhooks are triggered every time {props.description}.</p>
+      <VStack spacing={4} divide={true}>
+        {props.list.length === 0 ? <p className="text-muted">There aren’t any &quot;{props.title.toLowerCase()}&quot; webhook yet.</p> : props.list}
+      </VStack>
+    </div>
+  )
+}
+
+export default class ManageWebhooksPage extends AdminBasePage<ManageWebhooksPageProps, ManageWebhooksPageState> {
```
goenning, line 48: same here, if you could cover to function components please (see GeneralSettingsPage as an example to replace class inheritance)

public/services/actions/webhook.ts
```diff
@@ -0,0 +1,43 @@
+import { http, Result, StringObject } from "@fider/services"
+import { HttpHeaders, WebhookPreviewResult, WebhookStatus, WebhookTriggerResult, WebhookType } from "@fider/models"
+
+export const createWebhook = async (
```
goenning, line 4: because of the long list of params here, would it be worth having a type for it?

public/services/utils.ts
```diff
@@ -111,3 +111,11 @@ export const truncate = (input: string, maxLength: number): string => {
   }
   return input
 }
+
+export type Optional<T> = {
```
goenning, line 115: insn't this the same as `Partial<T>` which comes by default with typescript?

goenning (changes requested):

app/models/entity/webhook.go
```diff
@@ -0,0 +1,60 @@
...
+// Webhook represents a webhook
+type Webhook struct {
+	ID                    int                `json:"id" db:"id"`
+	Name                  string             `json:"name" db:"name"`
+	Type                  enum.WebhookType   `json:"type" db:"type"`
+	Status                enum.WebhookStatus `json:"status" db:"status"`
+	Url                   string             `json:"url" db:"url"`
+	Content               string             `json:"content" db:"content"`
+	HttpMethod            string             `json:"http_method" db:"http_method"`
+	AdditionalHttpHeaders HttpHeaders        `json:"additional_http_headers" db:"additional_http_headers"`
```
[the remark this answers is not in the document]
indyteo: Initially, I was thinking of a global configuration, then additional per Webhook, but I changed my mind on the way and this name remained unchanged, but you're right I'll rename it it'll be finer :)

app/services/sqlstore/postgres/webhook.go
```diff
@@ -0,0 +1,138 @@
...
+	"github.com/getfider/fider/app/models/entity"
+	"github.com/getfider/fider/app/models/enum"
+	"github.com/getfider/fider/app/models/query"
+	"github.com/getfider/fider/app/pkg/dbx"
+)
+
+func getWebhook(ctx context.Context, q *query.GetWebhook) error {
+	return using(ctx, func(trx *dbx.Trx, tenant *entity.Tenant, user *entity.User) error {
+		webhook := &entity.Webhook{}
+		if tenant != nil {
```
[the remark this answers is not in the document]
indyteo: Hmm yes you're probably right, but this is my trend to always check values when I'm not sure, even if it's not really necessary, I'll remove it

app/services/sqlstore/postgres/webhook.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+func listActiveWebhooksByType(ctx context.Context, q *query.ListActiveWebhooksByType) error {
+	return using(ctx, func(trx *dbx.Trx, tenant *entity.Tenant, user *entity.User) error {
+		webhooks := []*entity.Webhook{}
+		if tenant != nil {
+			err := trx.Select(&webhooks, `
+			SELECT id, name, type, status, url, content, http_method, additional_http_headers 
+			FROM webhooks 
+			WHERE tenant_id = $1 AND type = $2 AND status = 1
```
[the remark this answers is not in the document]
indyteo: Indeed, it'll be cleaner

app/services/webhook/props.go
indyteo: I didn't find any method to get a string representation, that's why I did mine, but it's true it would be better to add one instead!

app/services/webhook/webhook.go
indyteo: By testing you mean the "Trigger" button in the admin panel I suppose? Maybe `TestWebhook` would be confusing with unit tests, so `DummyTriggerWebhook`. Even if I don't feel very convinced by this name... Because in fact this is really triggering the Webhook, just with dummy props, not only testing the templates and URL validity for example.
I agree about the rename for the other command

app/services/webhook/webhook.go
indyteo: Yes, it's the point. An extra task here would be to notify the admins by email, but I didn't find the time to do it, and with the internationalization going, emails were being converted to translated templates at the time and I didn't risked myself into this...
But in theory, when a Webhook has failed, there's a reason behind this and I found this solution instead of keeping failing logs in an extra DB table (but maybe it would be better to do this, and thus being able in the admin panel to see the real failed request instead of a manually triggered one. It would make sense, what are your thought about this?).

app/services/webhook/props.go
indyteo: Interesting, I didn't noticed it was on the `AvatarType` enum. I'll implement it on the `Role` enum as well so :)

migrations/202107061359_create_webhooks.sql
indyteo: Damn, that's a lot of inconsistency... I must be tired at the time I did this 😆

app/services/webhook/props.go
indyteo: Concerning the single URL build, I think it would be counter-productive regarding the dummy props generation. Maybe we should add a `Url(Context)` method on `Post` to build the post URL in the given context?

indyteo: And for the params there are whether to include a complete post definition and the post author. I agree the names aren't well chosen, but a `complete` post definition include data that cannot be defined on post creation (votes, tags, response, etc...)

public/assets/styles/utility/display.scss
indyteo: Oh, what a good news :D

public/components/common/Loader.tsx
indyteo: Hmm I need to double check this, there may be a reason why I did it like that, or maybe adapt the Loader's SCSS to target the inner div (maybe a `c-loader__spinner` and a `c-loader__text`?)

public/pages/Administration/components/webhook/WebhookForm.tsx
indyteo: Yep, no problem I understand. Good to use those (no so) new React features

public/pages/Administration/components/webhook/WebhookForm.tsx
indyteo: It will be done

public/pages/Administration/components/webhook/WebhookProperties.scss
indyteo: I searched it at the moment, but didn't spotted it, I'll switch to it

public/pages/Administration/pages/ManageWebhooks.page.tsx
indyteo: I'll check this to help me make this inheritance, I have no idea about what to do it right now :o
By the way, shouldn't we deprecate the elements we're trying to move away? It may avoid theses troubles next time

public/services/actions/webhook.ts
indyteo: I think it was the reason why I made an extra (maybe already existing) TypeScript type definition, but I kinda messed up with it to be honest. I'll try to improve this little issue

public/services/utils.ts
indyteo: Looking at the `Partial<T>` type definition, it seems it is in fact exactly the same. Happy to know I wasn't the only one to do this, I suppose? 😆 (I really didn't know so much about TypeScript, and sometime I find easier to do things by myself, but when it already exist, it's good to use it!)

indyteo:
Finally, about the `webhook.Props` it's a really good idea, I like it! It's much more cleaner, I'll adapt the code to do so!

migrations/202107061359_create_webhooks.sql
indyteo: I'll change it to a `TEXT` type because it's a template and it may be larger than we expect (with route or GET parameters defined from the Go template). I've adapted the validation to handle the parsing before calling URL validation

app/services/webhook/webhook.go
goenning: Which makes me think that maybe the button should be "Test", just like the OAUth one? We test a webhook by triggering it with dummy props, but the action itself is still "Test" in my view.

app/services/webhook/webhook.go
goenning: Yeah, it's a lot more complex and we can implement that separately 👍

An email is probably a good starting point, otherwise webhook will get disabled silently. We can add this in a following PR

app/services/webhook/webhook.go
indyteo: I renamed those elements but kept the `WebhookTriggerResult` because it would be used by the hypothetical execution logging system

indyteo:
I had to do some refactoring to deal with cyclic imports, but I think I went on all your reviews! 😊

app/actions/webhook.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+// Validate if current model is valid
+func (action *CreateEditWebhook) Validate(ctx context.Context, _ *entity.User) *validate.Result {
+	result := validate.Success()
+
+	if action.Name == "" {
+		result.AddFieldFailure("name", "Name is required.")
+	} else if len(action.Name) > 60 {
+		result.AddFieldFailure("name", "Name must have less than 100 characters.")
```
goenning, line 34: 60 :)

goenning (commented):
We're nearly there! :D 

1. One validation error mentioned above.
2. I've been playing with different parameters for the body and noticed that `post_status` is a string and `post_old_status` is an integer. I'd expect both to be the same type and string is probably better.

On a similar note, it's not possible to quote an integer, so `{{ quote .post_number }}` doesn't work. What if we change quote to force conversion to string? So that `"number": {{ .post_number }}` would result in a number and `"number": {{ quote .post_number }}` would be a string.

I think these two properties add more noise and won't be useful, we could remove it.
- tenant_invitation
- tenant_welcome_message

3. Content outside the modal, do you see the same? Tested on Safari (Mac) and Chrome (Mac).
image

app/actions/webhook.go
indyteo: Damn I must be super tired to miss that 😆

indyteo:
Ok, I fixed the missing post status `.Name()` calls, removed useless tenants props and ensured the `quote` function properly en-quote the value. I also noticed there were no way to just escape a string (without adding extra quotes), so I added the `escape` function (example: `"title": "New post: {{ escape .post_title }}"`)

About the overflowing modal, on Firefox it's clean, I tested it on Safari Mobile (on my iPad), and it doesn't seemed to be overflowing (except when reducing the window to a very small size, so there may be an issue with responsive here, I think I found a workaround). I replaced all the HTML `<table>` because they were horrible when coming to responsive, please have a look now, it should be better :smile:

goenning:
Yes, it's a lot better now. The only issue outstanding is long values without space (such as URL). Maybe apply a `word-break: break-all;`. I can't think of anything else.

Screenshot 2021-08-17 at 19 41 54

After that we're good to go :)

indyteo:
I'm unable to reproduce the issue on my side (it get wrapped correctly on the next line), could you test if the `word-break` CSS rule do the trick for you?
I used to apply this CSS to break long words on other projects (I know the `hyphens` works well on Firefox but not on other browsers):
```css
overflow-wrap: break-word;
hyphens: auto;
```
So if the `word-break` don't work, maybe try `overflow-wrap`

goenning (approved):

PR #981, opened 2021-07-21
feature: add ability to copy comment link

indyteo:
**Feature request:** [Ability to copy comments permalink](https://feedback.fider.io/posts/317/ability-to-copy-comments-permalink)

As mentioned in the post, it is inspired by GitHub's "Copy Link" feature on comments.

It first add a unique link for each comment, based on their ID.
For example, my reply on this post has the ID `#31419`, so the link would look like :
```
https://feedback.fider.io/posts/317/ability-to-copy-comments-permalink#comment-31419
```
Then, it add the corresponding HTML `id` attribute on a wrapping `div` around comments, so the page automatically scroll to the targeted comment.
On the other hand, it also create a `highlighted` state in the `ShowComment` component that slightly mutate the comment background (a bit darker than others) to stand out. When a click is performed outside of the comment's area, it get back to normal and the hash is removed from the URL (the part after the `#`).

Highlighted comment

Finally, an action in the comment's drop down, target the current comment (set the URL in the navigation bar, triggering the comment highlight and scroll), and copy that link into the user's clipboard. A feedback indicate whether the copy was successful or not (due to the lack of support of the Clipboard API by some old outdated browsers), and invite people to manually copy their page's URL if not.

Copy link button

If the hash don't match anything is simply get ignored, and if it looks like a comment reference but we're unable to find it, a notice is shown to the user, because it may happen when the original comment got deleted, or the URL was manually modified and is now invalid. It get instantly cleaned up to avoid further share of a partially invalid link.

***

I never really played with URL's hash before and found it funny to do this, this is why I decided to train with this idea, even if it only has 2 votes (which means none, except from the submitter and me 😆), and it only took me about an hour so it was fun 😄


almereyda:
@indyteo Would you like to rebase this branch on `main`, or merge it again?

Then we could try to find someone for a review. Eventually you can also request a review?

indyteo:
Hi @almereyda, I can have a look if it's not much work 👍

mattwoberts:
@indyteo Hi!

Few years later.... is there much demand for this, I'm happy to review it if you want to rebase / merge from main, although I appreciate it's probably not the highest thing on your priority list right now :)

indyteo:
Hi @mattwoberts,

Let's be honest, I completely forgot about this one! 😄

I'd be happy to rebase (and btw clean that awful git history with all those merge commits...) so you can review. After a (very) quick review of the work that was done, I guess it shouldn't cause much trouble to rebase.
I'll handle this tomorrow that's on my todo-list ☑️ 

Thanks

indyteo:
@mattwoberts yay! 🎉

This should be up-to-date, ready to review! 😊
The git history is definitely cleaner too 😄

mattwoberts:
Great, thanks :)

I'll take a look soon!

locale/en/client.json
```diff
@@ -127,6 +128,9 @@
   "page.pendingactivation.text": "We sent you a confirmation email with a link to activate your site.",
   "page.pendingactivation.text2": "Please check your inbox to activate it.",
   "page.pendingactivation.title": "Your account is pending activation",
+  "showpost.comment.copylink.error": "Could not copy comment link, please copy page URL",
+  "showpost.comment.copylink.success": "Successfully copied comment link to clipboard",
```
mattwoberts, line 132:
```suggestion
  "showpost.comment.copylink.success": "Comment link copied to clipboard",
```

locale/en/client.json
```diff
@@ -127,6 +128,9 @@
   "page.pendingactivation.text": "We sent you a confirmation email with a link to activate your site.",
   "page.pendingactivation.text2": "Please check your inbox to activate it.",
   "page.pendingactivation.title": "Your account is pending activation",
+  "showpost.comment.copylink.error": "Could not copy comment link, please copy page URL",
```
mattwoberts, line 131:
```suggestion
  "showpost.comment.copylink.error": "Failed to copy comment link, please copy page URL",
```

locale/en/client.json
```diff
@@ -127,6 +128,9 @@
...
+  "showpost.comment.copylink.success": "Successfully copied comment link to clipboard",
+  "showpost.comment.unknownhighlighted": "Unknown comment ID #{id}",
```
mattwoberts, line 133:
```suggestion
  "showpost.comment.unknownhighlighted": "Invalid comment ID #{id}",
```

mattwoberts (changes requested):
Awesome stuff - some minor copy edit suggestions but all looks good!

mattwoberts (approved):

mattwoberts:
In it comes :)

almereyda:
This is so nice and useful!

Looking forward to the next release.

indyteo:
@almereyda it just got released to v0.23.0 😄🎉

mattwoberts:
@indyteo You beat me to it ;)

PR #985, opened 2021-07-26
feature: make tags clickable

indyteo:
**Feature request:** [Make tags clickable](https://feedback.fider.io/posts/133/make-tags-clickable)

This is very simple, but useful to quickly see posts with the same tag as another.
Tags now have a very small opacity variation according to mouse hover (0.9 / 1).


goenning:
I haven’t tried this branch yet, but what happens when you click on a tag from inside the admin panel?

I would have thought the tag should only be clickable on the home page, and potentially the show post page. Any thoughts?

indyteo:
The tag is clickable as soon as a slug is available.
This means in the admin panel the tags in the list are clickable, but not the previews when editing / adding tags.

If you think it's better, we could disable the link in the admin panel and instead add a "view tagged posts" button next to the "edit" and "delete" ones.

I think it's important for the admins to quickly see if a tag is used when they manage them. Maybe adding the count of tagged posts also: "view N tagged post(s)"

indyteo:
I added a prop to control whether the tag is clickable or not, and a button on the "Manage Tags" admin control panel page.

I don't know why the tests failed, I never used the system you use here (e2e) :confused:

goenning:
Hey, I'm sorry for the delay in testing this change. I had a look at this locally and I seeing it live now I agree with your first suggestion of making it clickable after saving on the admin area too. It looks a lot more cleaner and consistent. 

Do you mind reverting to that and we get this merged?

The E2E have passed now, was just some GitHub action issue I think.

Thanks

indyteo:
No problem, I just need to add that `link` prop and remove the extra button and we're good :smile:

Issue #730, opened 2019-03-06
Migrate components under public/components to React Hooks [1st step]


goenning:
Fider is on the process of slowly migrating to a class-free UI code by using React Hooks. This is going to be a long journey, but we need to start from somewhere.

This is the first step where we'll migrate all of our shared components, which are much smaller/simpler than the page components.

If you're interested in learning React Hooks and TypeScript on a real project, then join us 😊 

Leave your name and which component you'd like to work on.

These components are still classes and can be migrated:

- [X] common/form/Checkbox.tsx (➡️ @Anass-Daoudi #784)
- [ ] common/form/ImageUploader.tsx 
- [ ] common/form/ImageViewer.tsx 
- [ ] common/form/MultiImageUploader.tsx
- [ ] common/form/RadioButton.tsx 
- [ ] common/form/Select.tsx
- [ ] common/Button.tsx
- [ ] common/DropDown.tsx
- [ ] common/Header.tsx
- [X] common/Hint.tsx (➡️ @chagriali #788)
- [x] common/Loader.tsx (➡️ @goenning #733)
- [x] common/Modal.tsx (➡️ @goenning #809)
- [x] common/SignInControl.tsx (➡️ @chagriali #789)
- [x] common/Toggle.tsx (➡️ @goenning #732)
- [x] SignInModal.tsx (➡️ @goenning #809)
- [ ] VoteCounter.tsx

Blocked
- ErrorBoundary.tsx (cannot be migrated as Hooks doesn't support componentDidCatch yet)

**NOTE 1:** You don't need to know any of these technologies in order to work on it, if you need any help, just drop a message on this issue and we'll do what we can to help you out.

**NOTE 2:** If you'd like to practice your Unit Testing skills with Jest, then you can also add unit tests on your PR.

PR #1002, opened 2021-08-29
refactor: SignIn to hooks

benjlevesque:
**Issue:** https://github.com/getfider/fider/issues/730 

Refactor the SignIn component to use hooks instead of class components


goenning:
thanks!

PR #1006, opened 2021-09-11
i18n: add Russian translations

RuscalWorld:
**Issue:** #575 

Translated all strings to Russian. Maybe it is not good time to do this, but I realized that I need this for my project, and I would like to share these translations with others. If some new strings will be added in the near future, I'm ready to help with translating them too.

#### And few words about Crowdin
I've wanted to upload these translations to Crowdin, but I haven't found a way to do this. It looks like [this](https://github.com/getfider/fider/tree/main/locale#source-strings) is the only way to do this, so people are not able to upload new translations.


goenning:
Awesome! Thanks for your contribution! 

I've added Russian to Crowdin and already synchronised your translations. For updates and new translations, please use Crowdin and I'll sync it to git with all other languages

PR #1008, opened 2021-09-12
i18n: add slovak translation

jozefrebjak:
This PR add slovak translation.


goenning:
The build is failing because the files are on `sk/en/...` instead of `sk/...`

jozefrebjak:
> The build is failing because the files are on `sk/en/...` instead of `sk/...`

I'll fix it.

Issue #1026, opened 2021-11-07
Limit post description length in webhooks


RuscalWorld:
I'm using webhooks in Fider to publish all posts to Discord using its webhook system. The message in Discord can't be longer than 4000 symbols, but Fider doesn't have such limitation for post description. So, because of some users that create really big posts, some messages are not delivered to Discord webhook.

I think, there should be some property with short post description, so it can be passed to services that have limits on the message length. Maybe it will be better to provide an ability to reduce the post description to a certain length with template function like this:

```
{
	// ... some stuff
	"content": {{ shorten .post_description 4000 | quote }}
	// ... another stuff
}
```

```golang
"shorten": func(input string, length int) string {
	if len([]rune(input)) > length {
		input = string([]rune(input)[:length]) + "..."
	}
	return input
},
```

goenning:
Yeah, that sounds like a good approach. Do you want to submit a PR?

I would just suggest calling it "truncate"

RuscalWorld:
> Do you want to submit a PR?
Yes, sure. I will do this in the coming days

PR #1027, opened 2021-11-25
Implement "truncate" template function

RuscalWorld:
**Issue:** #1026

I've implemented `truncate` function that can be used in templates to shorten strings. It accepts *string* as first argument and *max length* as the second one. If given *string* is longer than *max length*, it truncates it and adds ellipsis at the end, so output string will have length that equal to *max length* argument.

This can be used to forward messages to services that have a limit on message length. For example, webhook for Discord will look like this:
```
{
    "content": "",
    "embeds": [
        {
            "title": {{ quote .post_title }},
            "type": "rich",
            "description": "{{ truncate (escape .post_description) 4000 }}",
            "url": {{ quote .post_url }},
            "author": {
                "name": {{ quote .author_name }},
                "icon_url": {{ quote .author_avatar }}
            }
        }
    ]
}
```

In this example `{{ truncate .post_description 4000 | quote }}` will not work because `quote` function escapes the post description and the final text will be longer than 4000 symbols if input text has special characters such as line breaks.

### DIFF
--- a/app/pkg/tpl/funcs.go
+++ b/app/pkg/tpl/funcs.go
@@ -66,6 +66,12 @@ var templateFunctions = map[string]interface{}{
 		quoted := quote(text)
 		return quoted[1 : len(quoted)-1]
 	},
+	"truncate": func(input string, length int) string {
+		if len([]rune(input)) > length {
+			input = string([]rune(input)[:length-3]) + "..."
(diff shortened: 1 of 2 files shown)


goenning:
Thank you! 🎉

PR #1035, opened 2021-12-09
Adding InsecureSkipVerify parameter

JianleiZhang:
client verifies the server's certificate chain and host name. Adding this parameter is convenient for users of self-signed certificates.


goenning:
There’s too many unrelated changes here, if you create a PR with only the code related to ignoring the certificate, I can merge

Issue #1047, opened 2022-01-24
[BUG] Docker container with HOST_DOMAIN is always unhealthy


iamjackg:
**Fider Cloud or Self Hosted**
Self hosted, 0.20

**Describe the bug**
The container's healthcheck uses `./fider ping`, which is hardcoded to call `localhost:3000`. If this doesn't match the provided `HOST_DOMAIN`, the check fails and the container is never marked as healthy.

**To Reproduce**
1. Start `getfider/fider:main`

**Expected behavior**
The container is marked as healthy when healthy.

goenning:
yeah, that's a good catch. can you try `adding -H "Host: $HOST_DOMAIN"` to your health check to see if that solves?

iamjackg:
Oh I'm good, thanks -- I removed the healthcheck altogether by adding the `disable:` key in my compose file since I'm not worried about it. It just caused me some pain because Traefik silently refuses to forward any traffic to a container if it's not healthy.

Just wanted to make you aware. Cheers!

gerkevgarderen:
Sadly, adding `-H "Host: $HOST_DOMAIN"` did not work for me. But I believe I found the cause and [created a PR](https://github.com/getfider/fider/pull/1048) with a fix.

goenning:
This should be fixed now with the latest `main` image.

PR #1048, opened 2022-01-31
Fix: Healthcheck errors if Host Domain set

gerkevgarderen:
Updated Ping command to fix health check errors

**Issue:** https://github.com/getfider/fider/issues/1047

When Host Domain is set, the Docker health check used to return errors.
This PR fixes that issue and also adds support for health checks if the app is running on ports other than the default 3000.

### DIFF
--- a/app/cmd/ping.go
+++ b/app/cmd/ping.go
@@ -11,6 +11,10 @@ import (
 //RunPing checks if Fider Server is running and is healthy
 //Returns an exitcode, 0 for OK and 1 for ERROR
 func RunPing() int {
+
+	client := &http.Client{}
+
+	host := "localhost:" + env.Config.Port
 	protocol := "http://"
 	if env.Config.TLS.Certificate != "" || env.Config.TLS.Automatic {
 		http.DefaultTransport.(*http.Transport).TLSClientConfig = &tls.Config{
@@ -19,12 +23,18 @@ func RunPing() int {
 		protocol = "https://"
 	}
 
-	host := "localhost:" + env.Config.Port
-	if env.IsSingleHostMode() && len(env.Config.HostDomain) > 0 {
-		host = env.Config.HostDomain
+	req, err := http.NewRequest("GET", protocol+host+"/_health", nil)
+	if err != nil {
+		fmt.Printf("Setting up request failed with: %s", err)
+		return 1
 	}
 
-	resp, err := http.Get(protocol + host + "/_health")
+	// Set Host if HostDomain is set
+	if env.IsSingleHostMode() && len(env.Config.HostDomain) > 0 {
+		req.Header.Set("Host", env.Config.HostDomain)
+	}
+
+	resp, err := client.Do(req)
 	if err != nil {
 		fmt.Printf("Request failed with: %s\n", err)
 		return 1
--- a/app/cmd/ping.go
+++ b/app/cmd/ping.go
@@ -31,7 +31,7 @@ func RunPing() int {
 
 	// Set Host if HostDomain is set
 	if env.IsSingleHostMode() && len(env.Config.HostDomain) > 0 {
-		req.Header.Set("Host", env.Config.HostDomain)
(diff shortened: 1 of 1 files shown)


goenning:
Ahh, we’re using a custom cmd for health check, I thought it was curl, so maybe that’s why it didn’t work.

the problem with this solution is that by sending a request to the actual domain, it’ll go through the internet instead of being a local request. So if the container is healthy but somewhere up the stack there is a problem, the health check will fail.

can you try adding a “Host” header to the http request? Curious to see if that’d work

gerkevgarderen:
> the problem with this solution is that by sending a request to the actual hostname, it’ll go through the internet instead of being a local request. So if the container is healthy but somewhere up the stack there is a problem, the health check will fail.

Yes, that's a fair point!

> can you try adding a “Host” header to the http request? Curious to see if that’d work

Sadly this does not work, for reference see the last commit.
I tried to test more in-depth by installing cURL in a running Fider container. Same result, the error originating from:
https://github.com/getfider/fider/blob/f9eed164d39fb865ce119753706e601827f7a204/app/middlewares/security.go#L22

Looking a bit more into that line, it seems the hostame in that check is originating from the actual Request URL. If I understand correctly, allowing the Host header to override this might mean allowing a bypass of this security measure.

Perhaps we can exempt the _health endpoint from the check? Not the most elegant solution but I believe this would not pose any significant risk as that endpoint does not send or process any information.

gerkevgarderen:
For reference, here's what I did for testing curl from inside the container:
- Start shell in container: `docker exec -it <FIDER_CONTAINER_NAME> /bin/bash`
- Install curl: `apt install curl`
- Run request with Host (also tested host for sanity):
  `curl localhost:3000/_health -H Host=<HOST_DOMAIN>`
- Try request with a resolve IP override:
  `curl --resolve <HOST_DOMAIN>:*:127.0.0.1 -I <HOST_DOMAIN>/_health`

goenning:
Thanks for the investigation above, I tried this `curl localhost:3000/_health -H Host=<HOST_DOMAIN>` and it also didn't work, but apparently the syntax is `-H "Host: value`, so when I tried:

`curl -k https://localhost:3000/_health -H "Host: <HOST_DOMAIN>"` I got 200 OK with `{}` response. So that worked in curl, but not in go.

So after some digging I found this https://github.com/golang/go/issues/29865

instead of `req.Header.Set("Host", env.Config.HostDomain)`, it has to be `req.Host = env.Config.HostDomain`.

---

> Looking a bit more into that line, it seems the hostame in that check is originating from the actual Request URL. If I understand correctly, allowing the Host header to override this might mean allowing a bypass of this security measure.

We had attacks before where requests where coming in as `https://{IP}/send-invite -H "Host: maliciousdomain.com`, before this check, we would accept the request and the `c.Request.URL.Hostname` was `maliciousdomain.com`. So every time we used that variable to build a URL and send by email, we were sending a hijacked url.

That's where this protection was introduced, if you override the Host header to a value that is different than what's configure, you get a 404.

goenning:
If that works for you too, we can merge this. By the way, good catch on the PORT, didn't notice we had it hardcoded :)

gerkevgarderen:
Ah, so both attempts to use the Host header were incorrect, sorry about that 😅 thanks for taking the time to investigate it too!

I haven't worked much with Go before so I thought checking with curl would help me ensure it wasn't my code but that didn't quite work out, haha.
I'll test your solution tomorrow and update the PR.

> That's where this protection was introduced, if you override the Host header to a value that is different than what's configure, you get a 404.

Ah okay, that makes sense. With your solution it seems we won't have to touch it, but this helps me to understand the setup better :)

Thanks for the fast response & engaging in finding the solution.

gerkevgarderen:
Just tested your fix, this works perfectly for our instance 👌

I have updated the PR with the fix, should be ready to merge.

goenning:
@gerkevgarderen thank you! 🚀

PR #839, opened 2020-01-15
Trust custom oauth providers in order to create unknown users in private mode

chrnin:
**Issue:**  https://github.com/getfider/fider/issues/838

I was wandering around, pretty bored, so I took a more deep look to this and tried something…
TLDR; here's a first try at trusting custom oauth providers that still needs work to be good… But this functionality fits exactly what I described in the issue.

what I did:
- This PR updates the database model to add is_trusted (bool) to oauth_providers table.
- In addition, this field is added in go models and oauth functions to take account of this new field.
- I also added a new option (always visible even if site is not private, maybe this could be improved to prevent bugging admins of public sites) in custom oauth providers screen.
- I manually tested that I was able to log-in in a private site with an unknown user on a trusted provider: success ! 
- I also manually tested that I was *not* able to sign in a private site with an unknown user on a *not* trusted provider: that failed and this is a success !

what I did'nt
- Unfortunately, my setup doesn't seem to take account of tests and `mage test` dies with a lot of garbage messages so I'm not able to run unit tests for the moment, that point needs further investigations :(
- I didn't test that pre-configured providers (google etc) are still working…

what I shouldn't have done
- My PR contains a huge diff for package-lock.json… I thought this file would have been ignored and didn't pay attention… My bad !
- I don't like much how I accessed customOAuthProvider in OAuthToken(), but as I understood it, this object is not loaded in this context at this moment… There are probably much more nice things to do.

Regards,

Christophe.

PR #1068, opened 2022-05-08
feature: trust custom oauth providers in order to create un-invited users in private mode

goenning:
Resuming work from #839


asos-robbell:
Is this included in a stable release now?

yannickdoteu:
Is there any way to mark the Google authentication as trusted? I have added it by updating the environment variables, but I cannot find a way to mark this method as 'trusted', so anybody from this OAuth method can access the site set as private.

goenning:
@yannickdoteu see https://github.com/getfider/fider/issues/1116

PR #1069, opened 2022-05-11
enhancement: replace HOST_DOMAIN with BASE_URL

goenning:
HOST_DOMAIN was introduced to avoid host attack, where an attacked could intercept the request, modify the Host header and Fider would send an email with a link using the hijacked Host domain. But this variable introduce multiple problems, for one, it's difficult to use in Kubernetes or behind any cloud load balancer.

This PR replaces it with BASE_URL, which is used when sending emails, but does not have any interference with the HTTP pipeline, meaning that Fider can be hosted on ANY url/port.


lcfd:
Hello 👋🏼 
I think that the doc at this [url](https://fider.io/docs/hosting-instance) should be patched. 
It still uses `BASE_URL: https://feedback.yoursite.com`.

Issue #1133, opened 2023-06-10
Scrollbars will always be visible in certain places like `VotesModal`


Ik1497:
![image](https://github.com/getfider/fider/assets/113038466/7ae3afe9-44a1-4cd9-bebd-7dd8a46039e3)

PR #1134, opened 2023-06-10
Scrollbars would always be visible in the `VotesModal` component

Ik1497:
**Issue:** All scrollbars would always be visible in the `VotesModal` component. See #1133


goenning:
Thank you :)

PR #1155, opened 2023-12-06
Add one-click deploy to readme

jbgh:
Hey!

We built this one-click deploy link for your users who may not know how to stand up their own infrastructure to easily self-host this.

After deployment, users receive a link like this, which can then be CNAME-ed to any domain:

https://fider-3074.dome.tools

Take it for a test run—we're eager to receive your feedback! We're more than willing to maintain this for the long haul, if this is valuable for your users.


mattwoberts:
Hope you don't mind me closing this - it doesn't look like dome has gotten much traction alas, and so I don't feel it's a suitable thing to add to the fider docs - please correct me if I'm wrong!

PR #1169, opened 2024-06-11
Added migration to create missing posts.user_id index

mway-niels:
**Issue:** 

As the number of users and/or posts grows, search performance goes down significantly (query duration >= 30s in our case).

Indexing the user_id column in the posts table fixes the issue.


mattwoberts:
Cheers @mway-niels ! Looks like a sensible index to have. I'll bring this in soon.

Issue #1143, opened 2023-07-23
[BUG] API List Posts default view is not all


darkdragon-001:
**Fider Cloud or Self Hosted**
_Couldn't find it in the UI or using the API_

**Describe the bug**
When calling endpoint [`/posts`](https://fider.io/docs/api#list-posts) without specifying `view` (`all`) it is not showing `completed` posts. When explicitly setting it to `all` they are included as well.

**To Reproduce**
1. Call API on `/posts`
2. Don't find any `completed` posts in the result

**Expected behavior**
Documented defaults should match reality.

**Screenshots**
_None_

**Additional context**
_None_

PR #1175, opened 2024-07-03
fix: setting default "view" query params to "all"

npham49:
**Issue:** https://github.com/getfider/fider/issues/1143

**Reasoning behind bug:** In the code for getViewData, by default statuses only contains "open", "started" and "planned", since  https://fider.io/docs/api#list-posts also uses the same function to generate the sql, the returned posts will only have "open", "started" and "planned" statuses. This conflicts with the documentations.

https://github.com/getfider/fider/blob/e55a144ab4aaeed1383bf0d9367cd80bfb7ec4a2/app/services/sqlstore/postgres/common.go#L33-L37

**Proposed changes:**
Since the getViewData function is used across the app, updating statuses here might cause problems on other parts of the app, I decided to create a conditional on the route handler for the API to always set the view Query Params to "all" if it is not specified.

### DIFF
--- a/app/handlers/apiv1/post.go
+++ b/app/handlers/apiv1/post.go
@@ -15,9 +15,13 @@ import (
 // SearchPosts return existing posts based on search criteria
 func SearchPosts() web.HandlerFunc {
 	return func(c *web.Context) error {
+		viewQueryParams := c.QueryParam("view")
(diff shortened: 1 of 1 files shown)


mattwoberts:
Thanks very much @npham49 :)

PR #1208, opened 2024-09-19
First set of proposed UI tweaks.

mattwoberts:
Fider's looking a bit dated, especially when you consider what else is out there.

So with that in mind we've been through the app and made some changes to the UI

- Inter is the main font for the site
- The main page and posts pages are changed to have a grey background and the main areas are boxed in
- "Action" area of the posts page moved to the left, so that it matches the main page, where the suggestion area is on the left
- Total redevelopment of the notifications - the old page is gone, and replaced with a much nicer (in our opinion) modal list of notifications
- Webhooks page changed to list all webhooks in one single list
- General spacing changes to let everything "breathe"


oddstr13:
I would prefer the actions stay on the right side, as that is in line with the GitHub UX.

I also generally dislike the "wasted space" from phone-style UI design, and personally prefer a more information dense display.

I have **not** spun up this pr to see how it looks, however, so take my comments with a small grain of salt :wink: 

Good to see the project moving again! :slightly_smiling_face:

mattwoberts:
@oddstr13 Thanks for the input, much appreciated.

Actions on the left - sure, it's a slightly subjective one, it does seem to work though so we'll see how it goes. If feedback isn't so good then we'll reconsider this one!

Issue #1211, opened 2024-09-24
Allow visitors to create posts with tags


MercierMateo:
Posts can be sorted with tags, but theses tags can only be added and modified by administrators and collaborators.

Users (with visitor role) might already know the necessary tags. It would make administration with tags a lot easier if visitors could specify tags while creating posts.

It would also help with automation. For example to send an email or a message to different contacts depending on which tag is used.

### Front-End

The post creation form could be extended to look like this model : 

![screen fider](https://github.com/user-attachments/assets/37c19a24-1eeb-4ec6-b280-6b9dcb1ed919)

This new field could be automatically filled with the tags specified in the URL through the tag filter search or an external link. It would help placing links on different web apps or physically through QR codes.

### API

Should the tagging and untagging permissions be given to the creator of the post after creation ?

- If this is the case the tagging and untagging routes would have to be made accessible to the creator :

#### (Un)Tag a Post

- Authentication: Required
- Required Role: Collaborator or Administrator
- Allow post creator to use
> Some other routes (like post editing) allow the post's creators to use them but it is not documented 

```
POST /api/v1/posts/{number}/tags/{slug}
```
...

- Else in the route used to create posts `POST /api/v1/posts` the payload could be extended to include tags :

#### Create a Post

- Authentication: Required
```
POST /api/v1/posts
```

##### Parameters
| Name | Type | Description |
| --- | --- | --- |
| title | string | Required. The title of the post. |
| description | string | The description of the post. |
| tags | list | The list of tags to create the post with |

##### Example

```
POST <baseURL>/api/v1/posts
---
{
  "title": "Allow users to edit any comment",
  "description": "This is useful to fix some minor typos or introduce more details"
  "tags": ["feature"]
}
```

Both these solutions would allow visitors to specify tags while creating posts.

### Feature enabling

Should this feature be behind a feature flag ? Or should it be enabled for everybody ?

MercierCorentin:
This feature request is quite important for us as we use Fider as a global feedback plateforme for our university projects, with @MercierMateo with your approval, we are ready to develop this feature ! 

It is part of an educational projet, I'm supervising the project.

mattwoberts:
Hey there.

Thanks very much for this. What you're proposing sounds like a nice addition to the app. 

We're more than happy for you to take this on, here are some thoughts / comments from us:

### UI

[...]

PR #1221, opened 2024-11-18
Allow visitors to create posts with tags

MercierMateo:
**Issue:** https://github.com/getfider/fider/issues/1211

Visitors can now create post with tags


app/actions/post.go
```diff
@@ -21,12 +21,39 @@ import (
...
+		getTag := &query.GetTagBySlug{Slug: slug}
+		if err := bus.Dispatch(ctx, getTag); err != nil {
+			return err
+		}
+		
+		input.Tags[i] = getTag.Result
+	}
+	
+	return nil
 }
```
mattwoberts, line 43: Good idea to make this part of OnPreExecute. You'll just need to update that comment though.

mattwoberts (approved):
Looking really good - apologies for the slow response in looking at this.

app/pkg/env/env.go
```diff
@@ -42,12 +42,13 @@ type config struct {
...
-	BaseURL    string `env:"BASE_URL"`
-	Locale     string `env:"LOCALE,default=en"`
-	JWTSecret  string `env:"JWT_SECRET,required"`
+	Port                        string `env:"PORT,default=3000"`
+	HostMode                    string `env:"HOST_MODE,default=single"`
+	HostDomain                  string `env:"HOST_DOMAIN"`
+	BaseURL                     string `env:"BASE_URL"`
+	Locale                      string `env:"LOCALE,default=en"`
+	JWTSecret                   string `env:"JWT_SECRET,required"`
+	PostCreationWithTagsEnabled bool   `env:"POST_CREATION_WITH_TAGS_ENABLED,default=false"`
```
mattwoberts, line 51: Yeah this looks good. I was contemplating if we should have a nested `AppFeatures` struct, and move this into there so that we could add other things in future, but I think we _probably_ don't need to do that...

mattwoberts:
Looking good. What are you thinking for tests?

MercierMateo:
On the action part I'm not sure anything is needed, but tell me if I'm wrong.

On the handler part I think the needed tests are : 
- create a post with an inexistent tag -> should fail as bad request
- create a post with a private tag as a standard user -> should fail as bad request
- create a post with a public tag as a standard user -> should succeed
- create a post with a public tag and a private tag as an administrator -> should succeed

As of now I think the fails are sent as a 404 error, so I will need to investigate on that before making these tests.

mattwoberts:
> On the action part I'm not sure anything is needed, but tell me if I'm wrong.
> 
> On the handler part I think the needed tests are :
> 
> * create a post with an inexistent tag -> should fail as bad request
> * create a post with a private tag as a standard user -> should fail as bad request
> * create a post with a public tag as a standard user -> should succeed
> * create a post with a public tag and a private tag as an administrator -> should succeed
> 
> As of now I think the fails are sent as a 404 error, so I will need to investigate on that before making these tests.

I suppose it depends - if you're handler tests poke the action then that's great, if they mock the action though then you might need some action tests to check the logic around building up the tags for example - see how you get on 👍

MercierMateo:
I created the tests, the actions are poked and function well.
Are these tests OK and enough for the backend part ?

app/handlers/apiv1/post_test.go
```diff
@@ -63,6 +64,185 @@ func TestCreatePostHandler_WithoutTitle(t *testing.T) {
 	Expect(code).Equals(http.StatusBadRequest)
 }
 
+func TestCreatePostHandler_WithInexistentTag(t *testing.T) {
```
mattwoberts, line 67:
```suggestion
func TestCreatePostHandler_WithNonExistentTag(t *testing.T) {
```

mattwoberts: Minor grammar change

mattwoberts (approved):
Cheers, yeah this is coming on nicely, there might be more tests as it progresses but no issues here

MercierMateo:
Thanks, i will look into the UI further but I struggle a bit more with this part.

I see there are errors on the ui test side, is it normal ? I don't think I made changes on the UI that could trigger these errors but I'm not sure.

mattwoberts:
> Thanks, i will look into the UI further but I struggle a bit more with this part.
> 
> I see there are errors on the ui test side, is it normal ? I don't think I made changes on the UI that could trigger these errors but I'm not sure.

Yeah it looks like something has changed, I think a new playwright version is causing problems for our github actions (https://github.com/microsoft/playwright/issues/30368) - it's nothing you've done. I'll take a look at this

mattwoberts:
@MercierMateo Ugh, something went badly wrong in github - I tried to add a commit to your PR and ended up merging it into main - sorry about this. I'll sort it out

mattwoberts:
@MercierMateo Ok confession time. I was trying out the `gh` command line tool for the first time, I made a change to your pull request, and thought I'd pushed it as a commit onto your fork. Turns out I had messed up badly, and pushed it to main, which resulted in your PR being merged into main and closed. I'd asked chatGPT how to do this and just blindly copied the command without properly checking it. Never trust ChatGPT.

What an idiot I am.

Sorry.

I see 2 options here

1) I revert all your commits from main to get it back to the state it was in before I did this. However I'm not sure what that would do for your pull request, and how we'd get the commits back into a pull request.

2) We just accept what I've done, and call that "phase 1" of your work. The functionality is working, you can create posts with tags, it's just that there is no way to do that from the UI. So we can get main working by fixing the issues with the build, and then you can create another pull request, for "phase 2" that then is the work to add this to the UI. 

What are your thoughts on this, and do you have any other suggestions?

Thanks!

Matt

MercierMateo:
Ok no problem, I'll add this to my list of reasons to avoid using ChatGPT

I think option 2 is fine. As I separated my work on the UI from my main branch, only the backend part was merged on your main. I can easily create a new PR and continue working on the UI part of the feature.

Obviously if there are any required fixes on the backend part we can go for option 1, but it doesn't seem like it.

mattwoberts:
Yes I think this is the best option - it's much easier all round I think, I've just got to sort out the issues causing the build to break on my side, and you can work on the UI.

mattwoberts:
Great stuff. I've fixed the issue with the tests failing now.

Matt

Issue #1228, opened 2024-12-03
Add support for version control hosting via Upsun.com


tylers-username:
With the addition of a `.upsun/config.yaml` and `.environment`, users who want to test out Fider quickly would be able deploy it to Upsun.com within a few short commands:

1. `git clone {fider}`
2. `upsun project:create`
3. `upsun push`

Upsun.com version-controls infrastructure, so Fider can define the default services and application requirements within the repository to reduce the setup burden of users who do not wish to deploy docker containers.

PR #1229, opened 2024-12-03
Add Upsun.com hosting support

tylers-username:
**Issue:** #1228

Add's version-controlled infrastructure to the Fider repository to enable developers exploring Fider to deploy to Upsun in a few short steps without needing to worry about docker.

Relevant documentation MR: https://github.com/getfider/fider-docs/pull/2

### DIFF
--- a/.environment
+++ b/.environment
@@ -1,0 +1,23 @@
+# Set database environment variables
+export DB_HOST="$POSTGRESQL_HOST"
+export DB_PORT="$POSTGRESQL_PORT"
+export DB_PATH="$POSTGRESQL_PATH"
+export DB_DATABASE="$DB_PATH"
+export DB_USERNAME="$POSTGRESQL_USERNAME"
+export DB_PASSWORD="$POSTGRESQL_PASSWORD"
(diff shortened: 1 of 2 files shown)


mattwoberts:
Hi there

Thanks for this, it's great to see Fider setup in different environments. However, I don't want to add upsun-specific files to the codebase, we want to support a variety of installation options, if we added these files it wold imply we favour upsun over other options, which isn't the case.

We've got a variety of different instructions for different platforms at [docs.fider.io](https://docs.fider.io/self-hosted), would you be willing to create a PR with instructions for that? Apologies, we're just in the middle of moving docs around so the repo is still a bit of mess at the moment, but that is certainly something we'd be happy to include as a self-hosting guide

Thanks
Matt

tylers-username:
Thank you. I'll provide instructions for the YAML files over at https://github.com/getfider/fider-docs/pull/2/files

tylers-username:
https://github.com/getfider/fider-docs/pull/2 has been updated to align with your suggestion @mattwoberts

PR #1236, opened 2024-12-29
Fix Arabic Selection Option

BasselArt:
**Issue:**


package.json
```diff
@@ -4,8 +4,10 @@
   "repository": "https://github.com/getfider/fider",
   "private": true,
   "dependencies": {
+    "@lingui/core": "^3.13.3",
```
mattwoberts, line 7: Why did you need to change these dependencies here?

public/assets/styles/variables/_text.scss
```diff
@@ -1,4 +1,4 @@
-$font-base: 'Inter', ui-sans-serif, system-ui, -apple-system, BlinkMacSystemFont, "Segoe UI", "Helvetica Neue", Arial, "Noto Sans", sans-serif,
+$font-base: 'IBM Plex Sans Arabic', ui-sans-serif, system-ui, -apple-system, BlinkMacSystemFont, "Segoe UI", "Helvetica Neue", Arial, "Noto Sans", sans-serif,
```
mattwoberts, line 1: We can't change the font, this change would make that font the default for all. We need to work out a solution that will choose Inter for all non-arabic systems, and use your font for all users who those the arabic language?

locale/ar/client.json
```diff
@@ -3,6 +3,7 @@
   "action.change": "تغيير",
   "action.close": "إغلاق",
   "action.confirm": "تأكيد",
+  "action.copylink": "نسخ الرابط",
```
mattwoberts, line 6: Sorry, we don't manage changes to the localised strings here.

Please can you make the changes to the arabic strings here: https://crowdin.com/project/fider

Then they can be pulled into the next release

Thanks

mattwoberts (changes requested):

app/pkg/letteravatar/letteravatar.go
```diff
@@ -0,0 +1,142 @@
+package letteravatar
```
mattwoberts, line 1: This looks interesting. So you use the font to generate the letter avatar, so that it supports arabic letters. That's cool - if we used this we'd just need to make sure we can support both arabic and non-arabic. Do you have any thoughts about how best to manage that?

package.json
BasselArt: I'm not developer, so I used Cursor to fix Arabic issue

public/assets/styles/variables/_text.scss
BasselArt: Agree with you, maybe some condition

locale/ar/client.json
BasselArt: Sure I will upload an updated version, but I still have some modification on it along with some missing strings

app/pkg/letteravatar/letteravatar.go
BasselArt: I give up and convert the Arabic letters to English,Ex: ب = B , from my very basic experience with php and converting text to image, Arabic language is very challenging

mattwoberts (commented):
Hi @BasselArt 

Thanks for the PR and the work you're doing to make Fider work in arabic.

I'm not sure however what i can do with this. From a quick scan through the code I see that many of the changes you have added break the existing app for other languages in order to make it work for Arabic. Obviously the only way that we could consider getting these pulled in would be if the changes were fully compatible with all languages.

Thanks

Matt

PR #1239, opened 2025-01-08
Update README.md selfhosting link fix

ShiftedMr:
**Issue:** 
I noticed the self-hosting link was broken while examining this software.
Figured I'd fix it :)


ShiftedMr:
Very welcome! Glad to assist. Figured something like that happened as those kind of redirects always slowly drift

Issue #1247, opened 2025-01-30
Add support for Resend.dev email client


mscno:
Description: Add an email service implementation for [Resend.com](https://resend.com/) to enable sending emails via their API. Ensure it follows the existing provider structure.

Expected Behavior: Users should be able to configure Resend as the email provider and send emails seamlessly.

mattwoberts:
Closing this as the PR was closed.

PR #1248, opened 2025-01-30
add support for sending emails with resend.com

mscno:
#1247 

This PR introduces support for [Resend.com](https://resend.com/) as an email provider, alongside existing AWS SES and Mailgun implementations. It allows users to configure Resend as their email service and send emails using Resend’s API.

Changes Introduced
	•	Added a new resend package implementing the email.Service interface.
	•	Integrated Resend Go SDK (github.com/resend/resend-go/v2) for email sending.
	•	Implemented sendMail function to send emails via Resend.
	•	Added basic tagging support (template and tenant tags) for tracking emails.
	•	Registered the service via bus.Register(Service{}), allowing dynamic configuration.
	•	Stubbed out fetchRecentSuppressions, as Resend does not provide an API for fetching suppressed addresses directly.

Implementation Details
	•	Uses resend.NewClient to initialize the Resend API client.
	•	Emails are sent with a subject, HTML body, and recipient list.
	•	Logs debug messages for email sending and failures.
	•	Handles missing sender/recipient addresses and ensures emails can be sent.

Next Steps
	•	Implement webhook-based suppression handling for tracking bounced emails.
	•	Improve error handling and add retries for transient failures.

Why This Change?

This change expands email provider support, allowing users to choose Resend as a more developer-friendly and modern alternative to AWS SES or Mailgun.


mattwoberts (approved):
Hey @mscno 

Many thanks. This looks good, can't spot any issues with it at all :)

> Implement webhook-based suppression handling for tracking bounced emails.

So in resend you need to keep track of bounced emails sent to a webhook? We can make it work that way, it's just a bit more work to track.

> Improve error handling and add retries for transient failures.

Yes a retry mechanism for all email sending would be quite good to have, agreed. It might be better to take that as a separate piece of work / PR though if we did look at that.

mscno:
Resend doesnt have a suppresion API like AWS so for now. 

So for now I just return nil and dont set any suppressed emails.

The only way to make it work is to manually listen to bounce webhooks from resend and then store these somehow. Probably this is a little too much work, for what its worth.

Its not strictly necessary to have the suppressed emails I think.


Agreed that the retry mechanism is probably better handled in the email service level rather than on the level of each provider.

mattwoberts:
> The only way to make it work is to manually listen to bounce webhooks from resend and then store these somehow. Probably this is a little too much work, for what its worth.

Yeah I think I agree, there's a lot of effort there to get the bounces to be processed like the others. I think we're OK to leave this.

mscno:
added here

https://github.com/getfider/fider-documentation/pull/1

kimtiago:
You must remove from the documentation as the PR wasn't merge

arantesxyz:
The docs got updated tricking people to believe `EMAIL_RESEND_API` is actually an option. LOL

mattwoberts:
There was a pr for it but it was closed, it’s probably quite easy to bring
it in I think.

But yes the docs were updated a little prematurely there, apologies!

mattwoberts:
Presumably @arantesxyz or @kimtiago you spotted this because you wanted resend support? If you want to sort this out - recreate this PR, check it's working for you, I'd be more than happy to bring it in...

Issue #1270, opened 2025-03-27
Faster selection of tags in edit mode


mattwoberts:
Here is the fider post:

https://feedback.fider.io/posts/557/faster-selection-of-tags-in-edit-mode

This is fairly well specced out - if anyone is happy to pick this up there is a bounty of 50 USD on completion (we will pay this from the open collective account)

abhishek818:
@mattwoberts will take this up

mattwoberts:
@abhishek818 Ok thanks! Let me know if you have any questions re: implementation or UI

mattwoberts:
Hi @abhishek818 - I think you might have got confused about what needs to change here sorry. It's when you are an admin and you edit a post - in there you can assign tags to a post - that is where we need the filter... You can look at the design screenshots from the fider feedback post to see the suggestions for how this can look..

npham49:
Hi @mattwoberts, if you don't mind having a look, I also have a suggested solution: https://github.com/getfider/fider/pull/1276

PR #1276, opened 2025-04-04
Making the TagsPanel a custom Multi-Select component

npham49:
Faster selection of tags in edit mode
**Issue:https://github.com/getfider/fider/issues/1270** 

Description of changes:
- Updating the TagsPanel component to display a Multi-Select Dropdown on edit, this is inspired by the Multi-Select dropdown screenshots shown in the Issue and the PoC in the ticket.
- TagsList can overflow when there are multiple tickets, adding custom CSS styling to make sure the tags fit within the container.

Reasoning:
- TagsPanel are only called within the ShowPost component and is responsible for assigning tags, make sense to edit this one.
- Modelled UI after screenshots and what is already available.

Possible improvements:
- Better use of SCSS variables
- UI tests
- Better use of shared components
- Each edit fires off an API request, to save cost it might make more sense to group it into one request. This will be a bigger change that will affect the backend

Demo:
https://github.com/user-attachments/assets/bcb099c6-163a-44c1-9402-2c00929be564

Images:
image


mattwoberts:
@npham49 Oh great, thanks I will take a look at this today!

re: docker - Did you look into seeing if you could solve that with config - is there something sub-optimal about the docker compose that could be improved I wonder for lower-memory machines? 🤔

mattwoberts:
@npham49 This is really nice, thanks very much - makes the tag editing so much nicer. 

Can I suggest some minor changes

- A little more padding to the right of the x in the tag
- Could you order by all private first, then all public
- Blank state "Select Options.." should be "Select Tags.." and "No options available" should be "No tags available"

But honestly, I'm really happpy you did this - it's made my day 😊 The $50 bounty is yours, I just need to work out how best to do it in opencollective !

mattwoberts:
Hey @npham49 - just on the docker compose thing you mentioned, I'm a little confused about this - I just checked and mine is using less than 400 Meg:

CleanShot 2025-04-04 at 15 35 30@2x

So I wonder what's happening there - are you on windows (Just wondering since docker makes more use of VMs in windows)

npham49:
@mattwoberts Thanks! I'm so glad you liked the solution, and more than happy I was able to help. I'll make those changes after work today and ping you when they're in. 

On the Docker, I'm on an M2 Macbook, using Rosetta Docker Desktop, so the performance might be a lot worst than Windows with the layers of emulation. Going to try and see if there is an alternative option for Docker Desktop!

npham49:
Hi @mattwoberts, I made the suggested edits you had. I also created a new utils function called sortTags that can be used across the app to make sure the tags would display consistently (private first and in alphabetical order!). I also added a test for this function.

npham49:
@mattwoberts my bad! Did not check the last commit though-roughly! I opted for the useLingui hook described here https://lingui.dev/tutorials/react#internationalization-outside-of-react.

```placeholder={assignedTags.length ? "" : t`Select Tags...`}```

Let me know how it goes on your end! 
image

**Comment by mattwoberts** [comment]
@npham49 That's perfect, thanks :) I'll get that merged in soon!

File: public/pages/ShowPost/components/TagsPanel.tsx [hunk 1]
```diff
@@ -43,33 +49,102 @@ export const TagsPanel = (props: TagsPanelProps) => {
...
+          </div>
+        ))}
+        {/* Search box to enter query string */}
+        <input
+          type="text"
+          value={query}
+          ref={inputRef}
+          onChange={(e) => setQuery(e.target.value)}
+          className="search-input"
+          placeholder={assignedTags.length ? "" : t`Select Tags...`}
```
mattwoberts, line 127: With this, how do you specify the code that you're going to use for "Select Tags..." ?

The way the codebase does this everywhere else is like this:

```javascript
i18n._("label.selecttags", { message: "Select Tags..." })
```

So you specify the code you're using. Would it make sense to do it that way?

mattwoberts (approved):
Looks great, just that comment about the lingui method to use?

public/pages/ShowPost/components/TagsPanel.tsx
npham49: It does make a lot more sense, I was trying to import i18n from the wrong package so ended up using "useLingui" from their docs. Make some edits to switch to i18n now!

mattwoberts (approved):

Issue #1290, opened 2025-04-19
Refactor to Support Dynamic RTL/LTR Locale Direction


Bilker1422:
### **Description:**

Currently, the application always uses **LTR (Left-to-Right)** direction, regardless of the selected language. This issue suggests refactoring the locale direction handling to dynamically adjust the text direction based on the language selected (e.g., **Arabic** should be **RTL**, while **English** should remain **LTR**).

---

### **Expected Behavior:**
- The **direction** of the text will automatically change based on the selected language:
  - **RTL** for languages like **Arabic**, **Hebrew**, etc.
  - **LTR** for languages like **English**, **French**, etc.
- Proper alignment and layout adjustments for **RTL** languages.
- The **dir="rtl"** attribute will be applied to the html element dynamically.

---

### **Steps to Reproduce:**
1. Change the language to **Arabic** (or another RTL language).
2. Observe that the layout and text direction still display as **LTR**.
3. The text and UI should adjust to **RTL** for languages like Arabic.

---

### **Additional Context:**
Currently, the text direction is statically set to **LTR**, and we need to refactor the code to check the locale/language dynamically and adjust the layout accordingly. This change is essential for proper **RTL** language support.

PR #1291, opened 2025-04-19
Refactor SCSS for RTL Support

Bilker1422:
**Issue:** [getfider/fider#1290](https://github.com/getfider/fider/issues/1290)

### What's this PR for?

This PR refactors SCSS styles to support right-to-left (RTL) languages and adds logic for setting the `dir` attributes on the `<html>` element based on locale settings.

### Changes:

- Injects locale-based attributes into the `<html>` tag:

  ```html
  <html lang="{{ .public.settings.locale }}" dir="{{ .public.settings.localeDirection }}">
  ```

- Replaces directional CSS properties with logical equivalents:

  ```scss
  padding-left   → padding-inline-start
  padding-right  → padding-inline-end
  margin-left    → margin-inline-start
  margin-right   → margin-inline-end
  border-left    → border-inline-start
  ```

- Adds RTL-specific styles for `.c-input`:

  ```scss
  :dir(rtl) .c-input {
    & + svg,
    & + img {
      left: 0;
      right: unset;
    }
  }
  ```


mattwoberts:
This is looking great, should be able to pull this in and get a new release out soon

Bilker1422:
@mattwoberts Feel free to reach out if you need support on any other features.

app/pkg/i18n/i18n.go
```diff
@@ -32,6 +33,10 @@ var localeToPlurals = map[string]string{
 	"zh-CN": "zh",
 }
 
+var rtlLocales = map[string]bool{
+    "ar": true,
+}
+
```
mattwoberts, line 39: Why is this a map, rather than a simple array - assuming we're just storing rtl langs here, this seems that it would simpler as an array (or slice)

Bilker1422: Map lookups have O(1) access time, while arrays or slices require O(n) to check for the existence of a value. Since I assumed we're performing this check on each page render, using a map would be more efficient. Correct me if I'm wrong.

mattwoberts:
Interesting point.

Disclaimer: I've inherited this project but Go is not my background, so I'm
not an expert on this. I'm actually not sure which one would be quicker, my
gut feeling was that an array would be faster for a small set. ChatGPT
seems to agree with this - especially if it's a very small set, but I
wonder if this is pretty academic, in reality it's not going to make a
noticeable difference

We _could_ benchmark to see if one is particularly faster in this scenario,
but I'm quite happy to stick with what you have to be honest.

Cheers.

Matt

PR #1240, opened 2025-01-12
Create post with tags from the UI

MercierMateo:
**Issue:** #1211
 
Follow up on this closed PR : #1221

Visitors can now create post with tags from the ui

PR #1306, opened 2025-05-28
Added support for tags to be set on post creation (#1211)

JimKnoxx:
I continued the work on this PR: https://github.com/getfider/fider/pull/1240 

In this PR I
- fixed the tests
- updated the dropdown UI
- improved spacing around the tag selection / image upload form fields
- added storing the selected tags in the cache 
- adding the current filtered tags (aka the query params) to the tag selection
  - when the user is send to fider with a preselected tag in the url (to filter only for a specific module / part) the user has the right tag already selected


public/pages/Home/components/PostInput.tsx
```diff
@@ -23,10 +26,19 @@ export const PostInput = (props: PostInputProps) => {
     return ""
   }
 
+  const getTagsCachedValue = (): string[] => {
+    const cacheValue = getCachedValue(CACHE_TAGS_KEY)
+    const urlValue = querystring.get("tags")
```
JimKnoxx, line 31: Need to check if this will do stuff even when the option to set tags is deactivated.

JimKnoxx: Fixed

JimKnoxx:
I implemented the other tags input in a slightly modified version. The tags can't be clicked to filter in this view and I changed the check if the user is allowed to edit tags.

Also added MercierMateo as a co-author, because I messed up the fork (we don't use forks where I work and I have to confess I didn't know how to do it correctly. I am sorry @MercierMateo, hope co-author is fine with you)

MercierMateo:
It's completly fine to me, I already didn't finish up my PR so it's great that someone did even a better job to add the feature.

Thanks for that.

public/pages/Home/components/TagsFilter.tsx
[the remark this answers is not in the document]
JimKnoxx: @mattwoberts I thing the PR is again ready to review.

mattwoberts (approved):
Great stuff, thanks @MercierMateo for starting this and for @JimKnoxx for completing it - it's a nice addition to Fider :)

Issue #1313, opened 2025-06-05
CSS styles are loaded multiple times


JimKnoxx:
I noticed in the browser inspector, that the css styles and color definitions are loaded multiple times.
Probably not a performance issue, but still not a good look.

For example:
![Image](https://github.com/user-attachments/assets/9ba3c385-5243-4558-bec3-f3bd3f4a8d8d)

mattwoberts:
Fixed (by merging your PR) in #1319 👍

PR #1314, opened 2025-06-05
Changed scss import rule to the use rule (#1313)

JimKnoxx:
In this PR the global variables.scss is handled like a module. This will result in less bloating in the browser inspector.

Also refactored the loading point for the color - themes.


mattwoberts:
Awesome, I'm not sure how the change removes the inspector bloat (that was really annoying me too), since I don't know sass too well, but this is great.

JimKnoxx:
I guess, that every import added that imported file to the global loaded style list. Probably thats why import was deprecated. use rule handles the styles in a local namespace, so only the component that loads the file has the styles. 

But thats only what I remembered from talking to a co worker about it. So might not be 100% correct.

mattwoberts:
Superseded by #1319

Issue #1333, opened 2025-06-24
[BUG] Layout issues


JimKnoxx:
**Version:**
Local Dev from current Main branch.

I noticed some layout issues with strings that have no spaces (eg. url etc) 

**Screenshots**

![Image](https://github.com/user-attachments/assets/df321028-d525-4ec6-b755-b43604df128f)

![Image](https://github.com/user-attachments/assets/9489b3e4-4b0e-4e2c-9191-081d0dda4740)

![Image](https://github.com/user-attachments/assets/778b8b93-9c86-4d7d-b15a-dbcc0335f60a)

PR #1334, opened 2025-06-24
Improved handling of long strings in titles and code blocks (#1333)

JimKnoxx:
**Issue:** https://github.com/getfider/fider/issues/1333

- I added a word-break class to allow easy breaking of long titles etc.
- I noticed that the spacing and sizing scss files where wrongly named and fixed that.
- Code blocks now have scrolling to prevent horizontal page scrolling.
- Comments have a calculated max width to prevent breaking the layout.

Also added:
- Better line breaking in mails

### DIFF
--- a/public/assets/styles/variables/_sizing.scss
+++ b/public/assets/styles/variables/_sizing.scss
@@ -1,5 +1,5 @@
-$spacing-inc: 4px;
+$sizing-inc: 4px;
 
-@function spacing($i) {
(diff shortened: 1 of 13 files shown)


public/assets/styles/variables/_sizing.scss
mattwoberts, line 1: Good spot!

mattwoberts (approved):
Makes sense, thanks for this

PR #1346, opened 2025-07-21
A11Y and filter improvements

JimKnoxx:
Hey there!
This time my changes are all in one PR :) But I tried to make them better to understand with own commits for each change set.

1)
So I noticed some parts with accessibility problems and I started work on them. 
- Some text to background contrasts weren't high enough (have to be >4.5)
- The "similar posts" on the new idea creation form were tab able even when no similar posts are listed
- The vote counter was only differentiated by color, if voted or not. So color blind people couldn't see if they voted. I improved that.

I will look into improving the color scheme a bit more to help with contrast values. Please feedback if I should do that in this PR.

2)
I also changed the "create post" button appearing out of no where to a simple disable/enable state. I would build this back if you hang on it a lot, but from a a11y standpoint it's not super great. I improved the validation though :D That I would keep.

3)
I added 2 more filter. "My Posts" and "No tags". To further improve the filter dropdown I added categories. "Own", "Admin", "Status" and "Tags". The "No tags" filter is sorted in the "admin" category is only for admin and collaborator visible. "My votes" and "my posts" is in the "own" category. Status and tags are self explanatory. 

I noticed some strange behaviour with the "my posts" filter, not sure if it's just my dev instance. Sometimes the value is not read correctly form the query parameter if I also added status filter. If you have an idea what happens here, I would be happy to fix it. 
Also my admin page seam a bit broken, have to recheck if this is by this PR or something else. So please check there too.


mattwoberts:
Oh wow, lots to go through! I'll take a look and report back.

Just on the disabled "create post" button - this was how it worked originally, but the issue we found was that when the modal opens, it could look like you can't create the post - "Hey the button is disabled, why am I not allowed to create ideas...." - so we tried a few things, and we liked the button appearing after you've entered enough content... Does that make sense?

JimKnoxx:
If it's too much for one PR I will split it up in 2 for you.

Okay, then I will reset my changes to the create button.

mattwoberts:
> If it's too much for one PR I will split it up in 2 for you.
> 
> Okay, then I will reset my changes to the create button.

I"m happy with one PR I think.

mattwoberts:
I've not been able to break the filter yet with My Posts selected... Also can't spot any issues in the admin section - some missing migrations? 🤔

JimKnoxx:
> My design co-founder might have some comments to make about the dropdown starting to look cluttered, but I think they're good changes, so happy to bring them in and if he has any suggestions I'll make them later.

Yea. Thats true. Maybe converting them into 2 dropdowns Status and Tags, add the "no tags" to the tags dropdown and put the personal filters into the user profile dropdown to be a bit hidden? They are probably not used very often? That would be a bit of a stretch though.

JimKnoxx:
> I've not been able to break the filter yet with My Posts selected... Also can't spot any issues in the admin section - some missing migrations? 🤔

Hmm, can be a missing migration. I haven't added any new rows to the sql files though? I'll look into that later again, maybe it was just a dev caching problem.

mattwoberts:
So I've chatted to my designer co-fouder, and here are our suggestions with the filter:

- "OWN" - rename to "MY ACTIVITY"
- Lose the Admin section, move "Without tags" to either the top or bottom of the tag list, and make visible to all. Also rename to "Untagged"
- We think it might look nicer without the divider line between the sections

And perhaps out of scope but some other things we discussed:

- Blank state - you shouldn't see the "Tags" section header if there are no tags
- Counts - just noticed the status filters have counts, why just those, should really have counts for everything there on the filter maybe - this might be well out of scope though!

public/components/common/form/CommentEditor.scss
```diff
@@ -127,7 +127,7 @@
 
     // Placeholder
     p.tiptap-is-empty:first-child::before {
-      color: var(--colors-gray-600);
+      color: var(--colors-gray-700);
```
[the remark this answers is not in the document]
JimKnoxx: > If we darken the placeholder here, we'll need to change it also for the normal text inputs - if you look at the new post modal now then the placeholder for the title is lighter

The a11y tools I use don't complain about the contrast on the post modal. Can you provide a screenshot of what you mean exactly?

JimKnoxx:
I have changed the filter to your suggestions. Also added the blank state you mentioned. For the counts I would need to look into it further, maybe I'll look into that in a later PR.

Also reverted the submit button, so it will appear again. 

Found some errors with the modal: When pressing esc in the tag selection, the modal would close. And when pressing enter in the title input the modal would also close. Both I prevented now. 

I noticed that sometimes closing the modal will send me back to the last opened idea, but I haven't found a quick way to fix that, I don't like messing with history 😄 (Open idea xzy -> homepage -> create idea -> esc -> (normal) homepage -> create idea -> esc -> back to the idea page of xyz) 

Also some more a11y fixes: the elements behind the modal where reachable by keyboard, I fixed that.

public/components/common/form/CommentEditor.scss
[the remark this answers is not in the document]
mattwoberts: > > If we darken the placeholder here, we'll need to change it also for the normal text inputs - if you look at the new post modal now then the placeholder for the title is lighter
> 
> The a11y tools I use don't complain about the contrast on the post modal. Can you provide a screenshot of what you mean exactly?

I just mean that in the share post page at the moment, the placeholder for the description (the editor) and the title are the same gray - in this PR you'd changed the placeholder color for the editor, but that means it looks different to the one for the title.

mattwoberts:
This is great, thanks!

 
> I noticed that sometimes closing the modal will send me back to the last opened idea, but I haven't found a quick way to fix that, I don't like messing with history 😄 (Open idea xzy -> homepage -> create idea -> esc -> (normal) homepage -> create idea -> esc -> back to the idea page of xyz)

Agh - I've seen this once or twice but not been able to repro, I'll take a look and see if I can get this to happen

@JimKnoxx Update: I've tried all sorts and I can't get this to happen at all... I'll keep trying, I'm using zen browser (basically firefox) so I'll try chrome

@JimKnoxx Update 2: Got it! I'll fix this :)

mattwoberts:
@JimKnoxx Would you mind sending me an email - I can't find yours and there are a few things I wanted to talk about

JimKnoxx:
Changed to draft to not be mergeable. Sign in is completely broken!

JimKnoxx:
Sorry about that draft/no draft. I should restart my dev setup more often.

public/pages/Home/components/PostFilter.tsx
```diff
@@ -76,7 +87,8 @@ export const PostFilter = (props: PostFilterProps) => {
   const options: OptionItem[] = []
 
   if (fider.session.isAuthenticated) {
-    options.push({ value: true, label: i18n._({ id: "home.postfilter.option.myvotes", message: "My Votes" }), type: "myVotes" })
+    options.push({ value: true, label: i18n._({ id: "home.postfilter.option.myvotes", message: "My votes" }), type: "myVotes" })
+    options.push({ value: true, label: i18n._({ id: "home.postfilter.option.myposts", message: "My posts" }), type: "myPosts" })
```
mattwoberts, line 91:
```suggestion
    options.push({ value: true, label: i18n._({ id: "home.postfilter.option.myvotes", message: "My Votes" }), type: "myVotes" })
    options.push({ value: true, label: i18n._({ id: "home.postfilter.option.myposts", message: "My Posts" }), type: "myPosts" })
```

(You'll need to change the strings that have been put into client.json too)

locale/en/client.json
```diff
@@ -157,6 +161,7 @@
   "newpost.modal.title": "Share your idea...",
   "newpost.modal.title.label": "Give your idea a title",
   "newpost.modal.title.placeholder": "Something short and snappy, sum it up in a few words",
+  "pDgeaz": "{title}",
```
mattwoberts, line 164: Any idea where this has crept in from? Looks like a `<Trans>` or `i18n._` call with a missing id, but I can't spot one on your commits?

JimKnoxx: Probably from this line: https://github.com/getfider/fider/pull/1346/files#diff-f2aeca8a64cb8fb86c02647741ce1af5b57caed930075dff5241b31b97eae24eR130

PR #1349, opened 2025-07-23
Fixed stacking up of escape handlers.

### LINKED ISSUE #1348: [BUG] Pressing ESC to close share feedback modal can land you on the wrong page.
As pointed out by @JimKnoxx , if you open the share feedback modal, and then press escape to close it  - fine. If you then do it again you can get too many history back events firing, which means you end. up on the wrong page.

mattwoberts:
**Issue:** Fixes #1348 

This fix means the modal component tidies up after itself properly to stop double event firing.


JimKnoxx:
This fixes the bug for me 👍

Issue #1368, opened 2025-10-02
[BUG] Long post titles break in the middle of a word when wrapping


widal001:
**Fider Cloud or Self Hosted**
Fider Cloud

**Describe the bug**
Long post titles wrap to multiple lines, but they often break in the middle because of [this SCSS](https://github.com/getfider/fider/blob/463ddfee7e77460646c91ad0d10ec0c0160c1dcd/public/assets/styles/utility/text.scss#L136-L139):

```
.text-break {
  word-break: break-all; # should be break-word
  overflow-wrap: break-word;
}
```

This is especially pronounced on mobile:

<img width="388" height="293" alt="Image" src="https://github.com/user-attachments/assets/8c367a26-3a68-4ba9-a412-cc912761a66a" />

**To Reproduce**
1. Go to [Demo.Fider.io](https://demo.fider.io)
2. Click on "Try it out here" to start a new post
3. Make a post with the title: "Super duper extremely long uninterrupted title that will probably wrap multiple lines"
4. You'll see the word "probably" broken in the middle ([sample post](https://demo.fider.io/posts/532/super-duper-extremely-long-uninterrupted-title-that-will-probably-wrap-multiple-lines))

<img width="777" height="439" alt="Image" src="https://github.com/user-attachments/assets/07f13e6d-8647-4460-bd58-6b78341ff815" />

**Expected behavior**

Lines should break around the word so "probably" appears on the next line

<img width="752" height="458" alt="Image" src="https://github.com/user-attachments/assets/3fe360f8-16bd-4727-a20b-823f06d59f7f" />

**Screenshots**

Long titles breaking in the middle of words:

<img width="388" height="293" alt="Image" src="https://github.com/user-attachments/assets/8c367a26-3a68-4ba9-a412-cc912761a66a" />

Expected wrapping behavior, breaking around words

<img width="382" height="282" alt="Image" src="https://github.com/user-attachments/assets/23f63ead-cc87-45d2-a830-20cef5dd89e9" />

**Additional context**

Individual users can fix this by putting the following in their "Custom CSS section, but this should probably be the default:

```css
.text-break {
  word-break: break-word;
  overflow-wrap: break-word;
}
```

I have a PR with a one-line fix for this that I'll open after submitting this issue to reference in the PR.

mattwoberts:
Sorted and released :)

PR #1369, opened 2025-10-02
fix(style): How long post titles break when wrapping

widal001:
### Summary

**Issue:** #1368 

The CSS for the `.text-break` class would cause long post titles to break in the middle of a word when wrapping to the next line. This changes the default behavior to break around the word instead.

As far as I could tell this class was only referenced by the post title in the list page and the post page:

```
grep -Hnr "text-break" . --exclude-dir ./dist | cut -d: -f1,2
./public/assets/styles/utility/text.scss:136
./public/pages/Home/components/ListPosts.tsx:27
./public/pages/ShowPost/ShowPost.page.tsx:220
```

### Current behavior

Screenshot 2025-10-02 at 4 35 24 PM

### New behavior

Screenshot 2025-10-02 at 4 35 49 PM

### DIFF
--- a/public/assets/styles/utility/text.scss
+++ b/public/assets/styles/utility/text.scss
@@ -134,6 +134,6 @@ pre:last-child {
 }
 
 .text-break {
-  word-break: break-all;
+  word-break: break-word;
   overflow-wrap: break-word;
 }


mattwoberts:
Thanks @widal001 , good spot I'll get that in asap!

PR #1373, opened 2025-10-08
Content moderation - WIP

mattwoberts:
WIP content moderation and commercial features branch...


JimKnoxx:
I was reading through the docs of this PR and wondered why this feature will be placed behind a paywall? Or am I understanding the licensing approach wrong? 
Does this also apply for self hosting? ATM the github readme states, that a selfhosted fider is 100% free.

mattwoberts:
> I was reading through the docs of this PR and wondered why this feature will be placed behind a paywall? Or am I understanding the licensing approach wrong? Does this also apply for self hosting? ATM the github readme states, that a selfhosted fider is 100% free.

We're probably going to move to a "open core" model (e.g. something like what cal.com do), where everything is free, but with an additional set of features that aren't free. If you are in "fider cloud", you have the app free, but to use the "commercial" features you need to pay (so it's like moving off the free tier). If you're self hosting, you'll need to get a license file which will enable you to use the commercial features on your self-hosted fider too. 

This is not concrete - if you have any thoughts on it then please let me know. If you think this would be a very bad move - be honest and let me know! Basically we're trying to come up with a model that works and allows us to make a living from building on and supporting the app. The "self host or pay for fider cloud" thing isn't working too well at the moment.

JimKnoxx:
Understandable reasons! 

As long as features are not going away to then become a pro feature I think this can work out. Giving specialty tools for more "professional" use of the software sounds legit, when sponsoring is not possible otherwise.

mattwoberts:
I'm going to close this for now - internall discussions about what we do about open core licensing are happening so this might not yet be the direction we go in....

Issue #1374, opened 2025-10-10
[BUG] Search / SimilarPost Results missing full matches


JimKnoxx:
**Fider Cloud or Self Hosted**
Self Hosted v0.29

**Describe the bug**
We got more and more user confusion, because people can't find their own posts when searching for them and noticing, they create duplicates, because the similar post feature does not find the posts with matching words.

**To Reproduce**
I created some posts in my dev fider (title comes from real requests)
- Abwesenheiten Vertretungsplan
- VertretungsBoard
- Vertretungsplanung Wochenansicht
- Vertretungsplan drucken
- Ungeklärte Vertretung wird zu Entfall
- Neuer Stunden/ Vertretungsplanwächter
- Wann kommt der "Wächter" für das neue Stunden- und Vertretungsplanmodul?

When searching for "Vertretung" I get only 3 results:
- Ungeklärte Vertretung wird zu Entfall
- VertretungsBoard
- Vertretungsplan drucken

Some word combinations are found, but most of them are not.

After looking into fixing it myself I found in [postgres/post.go 
](https://github.com/getfider/fider/blob/6b5f5aa163061f2d7933e407d1503477486302b1/app/services/sqlstore/postgres/post.go#L453) the scoring rules.

I found three problems: 
1: The scoreFiled is build for 'english' only, no matter what the tenent language is set to. Having english as a fallback is still a good idea, because it might be used in technical terms, in error logs posted in the description, non native speakers etc.
2: Also only `to_tsquery` is set to english, `to_tsvector` also needs a language context, otherwise these two will not necessary have the same context, see: https://stackoverflow.com/a/20462024 
3: `to_tsquery`will only match the word stem. In german we have connected words, "Vertretung" is a word stem but "Plan" is also a word stem. So it will stem "Vertretungsplanung" to "Vertretungsplan". (Why the search result found some connected words... sql magic I guess?). But this one was fixable for me, when I changed `to_tsquery` to allow prefix matches. `to_tsquery('german', $3 || ':*')`

If there is no backlash to this I will look into getting a MR ready with a fix.

mattwoberts:
Hey @JimKnoxx 

Good spots. It's surprising that the results are coming back so poorly for non-English (e.g. you'd think almost exact matches on title would work), but we're obviously getting it wrong in the psql functions you're talking about.

More than happy for you to take a look at this, thanks.

JimKnoxx:
Little update:
[...]

PR #1375, opened 2025-10-10
Fixed problems with search

JimKnoxx:
**Issue:** https://github.com/getfider/fider/issues/1374

Fixing search and similar post filter not finding exact matches.
Fixing filter "My Posts" showing all posts, if the user has no posts.


JimKnoxx:
I think this can be reviewed now. In my tests this works now for multiple different languages and uses all words. It still is not perfect but can actually find stuff now.  
I removed the filtering by rank score, all results are now shown, but still sorted by a rank calculation.

mattwoberts:
@JimKnoxx thanks, I'm not very well at the moment but will try to look as soon as possible.

JimKnoxx:
sure! Get well soon!

mattwoberts:
I also fed this into chatGPT (of course ;)) to see what words of wisdom it would generate - some half useful stuff in here

CleanShot 2025-10-15 at 21 06 14@2x

JimKnoxx:
No worries. I am new to postgresql search and build this mainly by this blog post, some stackoverflow resources and some AI help: 
- https://xata.io/blog/postgres-full-text-search-engine
- https://stackoverflow.com/a/21299033 and https://stackoverflow.com/a/20462024

1) In my sources they did combine the vectors for each language by an empty string and combine different languages by an "OR"

2) I will look into that, maybe it works without now. I mostly used it for bettter non latin support

3) Good point. I will try remove the unaccent again, maybe the stemming is working good enough now. The accent part gave me an headache tbh.  

4) I will ad a "AS score" or something.

5) I would love the get a metric on this one, but its not so simple. Is it possible to import posts and comments from an other instance? I see the backup option... Then I could test the changes on "real" date :) 

I need to look into creating indexes, haven't done something like this :)

JimKnoxx:
okay, seemingly the unaccent and ILIKE did not contribute to the results.

mattwoberts:
Thanks @JimKnoxx 

One thing I spotted and it's on my list to watch: https://robconery.com/postgres/creating-a-full-text-search-engine-in-postgresql-2022/

JimKnoxx:
I'll have a look at that!
When trying to use the websearch_to_tsquery variant, I noticed that it would only really work, if the `similarity` part got removed again (otherwise exclusion by minus rune wouldn't trigger) and then the search results would again be very bad.

mattwoberts:
Hey @JimKnoxx 

That video is great, and it really helped me understand this in more detail.

Main takeaways for me..

- We should def use a generated column for the tsvector
- Combining simple and <lang> does make sense from watching that video
-

JimKnoxx:
I tried adding a SQL migration, that adds the generated search column and a GIN, but I can't get it to work with the SQL logic in the post.go file. 

Also a generated column will only work with the data of the same table. We have to build the search tenant language dependent. So a auto generated column will not work. AI gave me a trigger based solution, but then, back to getting the search column into the SQL logic, without breaking everything else...

mattwoberts:
Hey again @JimKnoxx 

So I looked again at this just now, and I'm finding that websearch_to_tsquery is working quite nicely for me. I've also tested some of the other things from that video, and they also seem to be working well for me (ranking with both the simple and english, and also searching on both)...

Here's my current test SQL - I've simplified the search query to remove some of the noise - and I've moved the vectors into the fields to avoid a bit of repetition. 

```sql
SELECT
  *
FROM (       
        SELECT
          p.id,
          p.number,
          p.title,
          to_tsvector('english', p.title),
          to_tsvector('simple', p.title),
          p.slug,
          p.description,
          to_tsvector('english', p.description),
          p.created_at,
          p.status,
          to_tsvector('english', p.title) || ' ' || to_tsvector('english', p.description) || to_tsvector('simple', p.title) as vector,
          to_tsvector('english', p.title) || ' ' || to_tsvector('english', p.description) || to_tsvector('simple', p.title) || ' ' || to_tsvector('simple', p.description) as vector2         
        FROM
          posts p
          INNER JOIN users u ON u.id = p.user_id
            AND u.tenant_id = 1002
        LEFT JOIN users r ON r.id = p.response_user_id
          AND r.tenant_id = 1002
      LEFT JOIN posts d ON d.id = p.original_id
        AND d.tenant_id = 1002
  WHERE
    p.status != 6
    AND p.tenant_id = 1002
    AND p.status = ANY (ARRAY[0, 1, 4, 2, 3])) AS q
WHERE vector2 @@ websearch_to_tsquery('english', 'embedded')
or vector2 @@ websearch_to_tsquery('simple', 'embedded')
  --OR (similarity (title, 'embedded') > 0.3)
  --OR (similarity (description, 'embedded') > 0.3)
ORDER BY
  ts_rank(vector2, websearch_to_tsquery('english', 'embedded')) + ts_rank(vector2, websearch_to_tsquery('simple', 'embedded')) DESC
LIMIT 30
```

This is working well for my tests, but I'm just testing english at the moment, and I'm only just starting to poke around with it.

You'll see I have a "vector" and "vector2" - I was trying out different vectors, to see if it would be better / worse excluding description from the "simple" vector, I think "vector2" is doing better for my tests though.

The generated column / trigger is an interesting one though... I see why generated columns aren't going to work. I'm generally not a fan of trigger logic, since they tend to obfuscate logic away (although in this case it's very obviously just specific to generated the vectors so I think that's fine). However, we'd need to make sure that if we did create something, it would be possible to re-generate them all - if the user was to change language for example.

JimKnoxx:
I have some random russian words as a post (as non Latin test), some English posts and some German post copies from our productive instance. I noticed due to the live search that websearch didn't feel good. There were no results untill I had a complete word match, where the simple "to_tsquery" had the option with `:*` to match also the beginning part of a word. Thats why my to_tsquery [has the regex](https://github.com/getfider/fider/pull/1375/files#diff-c7f2e1e6da474d77fca0db95457b96e4bf7713e18847878726afde807363a3e4R406) to make each input word a prefix/wildcard match.

I'd love to have more posts to search through, to get a general feeling if one idea works better.

Also thought about a solution for the generated column. We could implement https://github.com/pemistahl/lingua-go

This could then calculate once the language of each post, fallback to current tenant.locale and we store this information on a new column for the post table itself. 
Then we can add a generated column for the search, as we now have all the informations about the language and wouldn't need a sub query. 

For each new post we assume the language is tenant.locale and just store this on the post. Changing locale would make new posts default to the new tenant language. 

I just don't know how to create a run once migration script, that would need some research for me.

How you feel about that?

mattwoberts:
@JimKnoxx 

I'm testing with an anonymised version of the fider cloud db. I'm finding it works well for me right now - including partially matching (propo works for proposals etc).. 

I could maybe get a copy of that SQL to you if that's useful...

Matt

mattwoberts:
@JimKnoxx NIce idea about the language detection - I was thinking also about bringing the locale into the post for the generated column, this adds an extra layer so that you can have posts that are written in different languages and it will all work in theory.. Yeah I like this.

Ok - I've not got my head around the migration script. You need to create a script that adds the locale and vector cols, and then iterates through all posts and populates the locale (I presume for the migration we'd just use the tenant locale?). Then it bulk calculates all the vectors, and you're good to go? I could have glossed over something here?

mattwoberts:
Hmm, just tested with your posts from your issue...

Here's my query:

```sql
SELECT
  *
FROM (       
        SELECT
          p.id,
          p.number,
          p.title,
          to_tsvector('german', p.title),
          to_tsvector('simple', p.title),
          p.slug,
          p.description,
          to_tsvector('german', p.description),
          p.created_at,
          p.status,
          to_tsvector('german', p.title) || ' ' || to_tsvector('german', p.description) || to_tsvector('simple', p.title) as vector,
          to_tsvector('german', p.title) || ' ' || to_tsvector('german', p.description) || to_tsvector('simple', p.title) || ' ' || to_tsvector('simple', p.description) as vector2         
        FROM
          posts p
          INNER JOIN users u ON u.id = p.user_id
            AND u.tenant_id = 1002
        LEFT JOIN users r ON r.id = p.response_user_id
          AND r.tenant_id = 1002
      LEFT JOIN posts d ON d.id = p.original_id
        AND d.tenant_id = 1002
  WHERE
    p.status != 6
    AND p.tenant_id = 1002
    AND p.status = ANY (ARRAY[0, 1, 4, 2, 3])) AS q
WHERE vector2 @@ websearch_to_tsquery('german', 'Vertretung')
or vector2 @@ websearch_to_tsquery('simple', 'Vertretung')
--OR (similarity (title, 'Vertretung') > 0.3)
--OR (similarity (description, 'Vertretung') > 0.3)
ORDER BY
  ts_rank(vector2, websearch_to_tsquery('german', 'Vertretung')) + ts_rank(vector2, websearch_to_tsquery('simple', 'Vertretung')) DESC
LIMIT 30
```

And it's not doing well, without the similarity stuff, only one result here. Why is this though, I don't understand? It's like it's not geared up for German very well.. 

For the title `Ungeklärte Vertretung wird zu Entfall` it comes back with these vectors:

`'entfall':5 'ungeklart':1 'vertret':2`

And the websearch_to_tsquery for `Vertretung` comes back with `'vertret'`

So that one one matches... but none of the others do...

This is I guess what you reported.. I'm just not convinced that adding the similarity is the solution here, there's something not right with the FTS...

JimKnoxx:
Maybe the similarity part helped out and masked the usefulness of the unaccent part I removed again?

Also stumbled over this blog post: https://www.dbi-services.com/blog/dealing-with-german-umlaute-in-postgresqls-full-text-search/ 
This might be overkill, as not many germans typing ä as ae (mostly programmers with an english keyboard)

mattwoberts:
I started adding a db test using the test titles you mentioned, I thought
this might be useful. I've not quite finished that yet but when I do it
might be useful, I"ll see if I can add a few more tests for it too.

I'm not totally against the "similarity" part, but to me it just seems a
bit of a red flag that the full text search isn't enough to find what
you're looking for...

mattwoberts:
@JimKnoxx Test added... I haven't bothered checking in my hacks with the query, I didn't want it to clash with anything you might be working on.

mattwoberts:
@JimKnoxx Take a look at my latest push and let me know what you think - more tests added, and the tweaked searchPosts seems to be doing pretty well (as you suggested websearch_to_query isn't doing too well), so this version seems to be performing pretty well... We still need to consider the performance - but from my testing so far this is running pretty fast... I think my next job would be to throw some against some live datasets to see how it does...

mattwoberts:
Note on the performance. Running this on the busiest hosted fider instance (about 1000 posts), which is a fairly modest spec hosted postgres instance, I'm getting about 800ms for a search. When I strip out all search stuff it's coming back at about 300ms.

app/services/sqlstore/postgres/post.go
```diff
@@ -487,8 +487,8 @@ func searchPosts(ctx context.Context, q *query.SearchPosts) error {
 
 			// Build tsquery with AND operator between words and prefix matching on each word
 			// regexp_replace adds ':*' to each word for prefix matching, then joins with ' & ' for AND
-			tsQueryExpr := fmt.Sprintf("to_tsquery('%s', regexp_replace(regexp_replace($3, '\\\\s+', ':* & ', 'g'), '$', ':*'))", tsConfig)
-			tsQuerySimple := "to_tsquery('simple', regexp_replace(regexp_replace($3, '\\\\s+', ':* & ', 'g'), '$', ':*'))"
+			tsQueryExpr := fmt.Sprintf("to_tsquery('%s', regexp_replace($3, '\\s+', ':* & ', 'g') || ':*')", tsConfig)
+			tsQuerySimple := "to_tsquery('simple', regexp_replace($3, '\\s+', ':* & ', 'g') || ':*')"
```
mattwoberts: There is a test for multiple words - what am I missing?

JimKnoxx:
Hey @mattwoberts 
Thanks for the changes, the idea to add a `:*` to get rid of the similarity is really good! 

I added this to the findSimilarPosts method too :)

Now we probably should look at building a search index on DB level. I probably will not have time for that today but will look at that this week.
You had a question about that:
> Ok - I've not got my head around the migration script. You need to create a script that adds the locale and vector cols, and then iterates through all posts and populates the locale (I presume for the migration we'd just use the tenant locale?). Then it bulk calculates all the vectors, and you're good to go? I could have glossed over something here?

My thought was, that we don't know, in witch language a post was written in the past. The current tenant language might not be this language. But in the end a basic "use the current tenant.locale" should work find, as the current search does decent, even for not tenant language words.

JimKnoxx:
Ok, I found some time but I have problems with this migration script... When I try to migrate I get the error
`function to_tsvector(text, character varying) does not exist` no matter what I try. Yes, the title filed is of type character varying but casting does not change anything... Google and AI are not helping. Do you have an idea what could help here?

JimKnoxx:
ok, why and how did it work in the pipeline... I had some funky migration script trial and errors, maybe my DB status is a bit wonky... *sigh*

JimKnoxx:
Ok, fixed. Migration script should work now. 
I also tested the language detection of new posts, that also seem to work. 
Just need to use the search column in the sql statement. 

Can you add this? You seem to understand this enormous sql construct a lot better then me.

mattwoberts:
@JimKnoxx I'll try to find to do this later on - cheers for what you've done :)

mattwoberts:
@JimKnoxx Apologies took me a while to get to this. All plumbed in, it's looking good - certainly working well for my tests, and it's faaast. How does it look for you?

JimKnoxx:
Just put the changes on my dev machine and everything works great! Thank you @mattwoberts for the help.

Do you want to stress test / benchmark it on a bigger system before merge?

mattwoberts:
@JimKnoxx I'm not sure... I feel like it's good enough to go. My only concern I suppose is how long this is going to take to run on all the existing posts on the fider cloud... it's > 50k posts on there... 🤔

JimKnoxx:
You don't have a maintenance mode I guess? 
Hard to tell, most of my posts only consists of a title for tests. 

We have about 800 posts on our live system atm, but I'm not allowed to play around with it for benchmarks.

mattwoberts:
Ha, I was just about to reply - I think I'd have to put it in maintenance mode to do this ;) 

Yes, I think that's the only thing I can do - should take about 10 mins I think, from a rough guess.

JimKnoxx:
Great! I am very much looking forward to the results of this change and hope everything will work well.

mattwoberts:
@JimKnoxx I spotted in the PR that the language handling code introduced a bit more fragmentation of how / where languages are defined and handled in fider, and since we keep getting requests for adding new ones, I felt like we needed to tidy this up a bit - so see previous commit, I've left the client side code as it is, but tidied the server side code so that all languages are defined in a central enum, which should make it a bit cleaner and easier to add new ones going forward.

JimKnoxx:
@mattwoberts That looks great. Having a single source of truth is really good for users. 

Just some input about the "step by step" tutorial to make it easier to follow through and to be found by people who wanna contribute :)

JimKnoxx:
Awesome! Very detailed and good to follow step by step instructions.

PR #1380, opened 2025-10-26
Fix: Use proper i18n for post status

kohe:
**Issue:** 

This PR fixes the post status display in the post table to use the proper i18n instead of hard-coded text (in English).
This ensures that post status labels are properly localized and always displayed in the user's language of choice.


mattwoberts:
Nice find, thanks for sorting.

PR #1384, opened 2025-10-27
Allow setting listen addr with HOST and METRICS_HOST

dani:
This PR adds support for configuring listen IP using the HOST (and METRICS_HOST) env var. Some environments need this to run securely (my use case is running fider on Nomad, and exposing it through the Consul service mesh, which requires service to bind on 127.0.0.1 to prevent random other allocation from bypassing defined ACL)


mattwoberts:
Cheers, I don't think I have any issues with this.

Issue #1446, opened 2026-01-29
Home filter: My Posts doesn’t apply on click


michaeltyson:
Fider Cloud or Self Hosted
Self Hosted — getfider/fider:stable (image ID 328c84f66930, created ~4 months ago)

Describe the bug
When clicking the “My Posts” filter on the Home page, the list reloads but shows all posts instead of only the current user’s posts.

To Reproduce
Sign in as a user with posts.
Go to the Home page.
Open the Filter dropdown.
Click “My Posts”.
Observe that results reload but show all posts.

Expected behavior
Only posts created by the current user should be shown when “My Posts” is enabled.

Screenshots
Not included.

Additional context
The browser URL updates to /?myposts=true, and a full page reload shows correct results. However, the XHR request after clicking the filter does not include myposts=true.

PR #1447, opened 2026-01-29
Fix PostFilter keys to prevent My Posts toggle collisions

michaeltyson:
**Issue:** #1446 

Fixes a collision in the Home filter dropdown where “My Votes” and “My Posts” shared the same key/field, causing the wrong option to toggle. Uses a unique key/field per filter type/value so “My Posts” applies correctly and the API request includes myposts=true.


mattwoberts:
I can't recreate the issue @michaeltyson but it makes sense what you've done, thanks!

PR #1449, opened 2026-02-04
Indexing is a pro feature.

mattwoberts:


JimKnoxx:
Hey there! 
Does this change make the public search indexing (google etc) a pro feature or the fider internal search index over all posts a pro feature?

mattwoberts:
@JimKnoxx It's the google indexing stuff.

So we used to get a lot of signups from spammers, expoiting us to get google indexing on their spammy posts. We solved this a while ago by disabling google indexing on new signups. Now that we've got the pro version, we've made sure that if you are on a pro account we turn this back on.

(Sorry for the slow response, I've got a lot on at the moment!)

JimKnoxx:
Thanks for the reply! 
Spammers are annoying, understandable change.

PR #1460, opened 2026-02-17
feat: Add configurable default post sort order

AZAA:
## Summary

Adds an admin setting in General Settings to configure the default post sort order (Trending, Most Wanted, Recent, Most Discussed).

Previously the frontend always fell back to "Trending" regardless of any server-side configuration. Now the tenant's `defaultSort` setting is respected both server-side (in the handler) and client-side (in the PostsContainer component).

## Changes

- New database migration: `default_sort` column on tenants table (defaults to `trending` to preserve existing behavior)
- Backend: entity, command, action, handler, and store updates to persist and read the setting
- Frontend: Default Sort dropdown in General Settings, fixed PostsContainer/PostsSort fallback

## Testing

- UI tests: 117/117 passing
- Manual testing: change default sort in admin settings, verify homepage respects it without `?view=` param


mattwoberts:
This might need a little more discussion - I get why this solves an issue, but I wonder if users might appreciate something specific to them - so we remember (via a cookie or something) what users choose and stick with that. Although - you could add both features, so admins can set the default and we let users set their own - something like that.

Related: https://feedback.fider.io/posts/457/allow-configuration-of-default-sort

mattwoberts:
(btw - where are my manners, thanks for the pull request - always appreciated!)

PR #1461, opened 2026-02-17
feat: Add toggleable Roadmap page with kanban view

AZAA:
## Summary

Adds a toggleable Roadmap page that administrators can enable/disable from General Settings.

When enabled:
- **Feedback** and **Roadmap** navigation links appear in the header with active state highlighting
- The Roadmap page displays posts grouped by status in a kanban-style view (Planned, In Progress, Completed)
- Each column shows post count, vote count, comment count, and supports pagination (10 posts per page)
- When disabled, the `/roadmap` route returns 404 and the navigation links are hidden

This PR also includes the default sort setting (adds `default_sort` column to tenants) since the Roadmap feature builds on the same entity/store changes.

## Changes

- Two database migrations: `default_sort` and `is_roadmap_enabled` columns on tenants table
- New `Roadmap()` handler with `GetPostsByStatuses` query
- New Roadmap page component with SCSS (responsive, 3-column grid on desktop, single column on mobile)
- Header navigation with conditional Feedback/Roadmap links
- Admin toggle and default sort dropdown in General Settings
- i18n: translation keys for menu and roadmap across all 20 locales

## Testing

- UI tests: 117/117 passing
- Manual testing: toggle roadmap on/off, verify navigation, verify kanban columns, pagination, responsive layout
- Manual testing: default sort setting respected on homepage


mattwoberts:
Hi @AZAA

Thanks for spending the time to do this.

However, it's a bit awkward, but this is something we're also working on internally. 

We're trying to get to a point where Fider is generating some income to allow us to spend more time working on it. We've recently made Fider free on the cloud version, with a pro version that offers more features. We're trying to add some features that are "pro" features that some users would be interested to pay to have. We decided that this roadmap feature would be a good candidate for that, so this is what we've been doing.

It means we're not going to accept yours though, sorry. 

I would suggest that if you're planning to work on any other new features in fider, perhaps if you open an issue and raise the feature first for discussion, it might have provided an opportunity for us to respond and let you know what the situation was?

Thanks

Matt

PR #1502, opened 2026-04-15
fix: Context.BaseURL() discards path when HOST_MODE=single

### LINKED ISSUE #1452: [BUG] `Context.BaseURL()` strips path from `BASE_URL`, breaking subfolder/sub-path hosting
**Fider Cloud or Self Hosted**
Self Hosted: 0.21.1

**Describe the bug**
When `BASE_URL` is configured with a path component (e.g., `BASE_URL=https://example.com/feedback`), the `Context.BaseURL()` method strips the path and returns only `https://example.com`. This breaks all redirects and frontend navigation when Fider is hosted under a subfolder behind a reverse proxy.

The root cause is that there are two `BaseURL` implementations that behave differently:

`Context.BaseURL()` (method on Context struct) — used by the renderer and handlers:
```go
// app/pkg/web/context.go:383
func (c *Context) BaseURL() string {
    return c.Request.BaseURL()  // always strips path
}
```

`web.BaseURL()` (package-level function) — NOT used by the renderer:
```go
// app/pkg/web/context.go:612
func BaseURL(ctx context.Context) string {
    if env.IsSingleHostMode() {
        return env.Config.BaseURL  // correctly returns full URL with path
    }
    ...
}
```

`Request.BaseURL()` only returns `scheme://host:port`, discarding the path:
```go
// app/pkg/web/request.go:105
func (r *Request) BaseURL() string {
    address := r.URL.Scheme + "://" + r.URL.Hostname()
    if r.URL.Port() != "" {
        address += ":" + r.URL.Port()
    }
    return address  // path is gone
}
```

The renderer passes `ctx.BaseURL()` (the context method) to the frontend as `Fider.settings.baseURL` at `app/pkg/web/renderer.go:215`, so the frontend also receives a `baseURL` with no path.

**To Reproduce**
1. Set `BASE_URL=https://example.com/feedback` and `HOST_MODE=single`
2. Configure a reverse proxy (e.g., nginx) to forward `/feedback/` to Fider, stripping the prefix
3. Open the Fider UI at `https://example.com/feedback/`
4. Perform any action that triggers a redirect (e.g., delete a post, sign in, OAuth callback)
5. Observe the browser navigates to `https://example.com/` instead of `https://example.com/feedback/`

**Expected behavior**
`Context.BaseURL()` should return the full `BASE_URL` including the path when running in single-host mode, matching the behavior of the package-level `web.BaseURL()` function. All redirects and frontend navigation should stay within the `/feedback/` sub-path.

**Additional context**
I'm using 0.21.1 of Fider. Self hosted behind a reverse proxy at a sub-path.

The impact is:
- `navigator.goHome()` redirects to `https://example.com/` instead of `https://example.com/feedback/`
- `navigator.goTo("/posts/1/my-post")` navigates to the wrong location
- Backend `c.Redirect(c.BaseURL())` sends users to the domain root
- All code relying on `c.BaseURL()` for redirect targets is broken

Proposed fix:
```go
func (c *Context) BaseURL() string {
    if env.IsSingleHostMode() {
        return env.Config.BaseURL
    }
    return c.Request.BaseURL()
}
```

This aligns the context method with the existing package-level `BaseURL()` function.

tejas-ae:
When `HOST_MODE=single` and `BASE_URL` contains a path component (e.g. `https://example.com/feedback`), `Context.BaseURL()` was returning only the scheme and host (`https://example.com`), silently dropping the path.

This broke server-side redirects (sign-in, OAuth callbacks, delete-post flows, etc.) and the `baseURL` value injected into the frontend via the renderer — all of which call `c.BaseURL()`. Users hosted behind a reverse proxy under a sub-path would land on the root of the site instead of the configured sub-path after any redirect.

The root cause is a split between two `BaseURL` implementations. The package-level `web.BaseURL(ctx)` already handled single-host mode correctly by returning `env.Config.BaseURL` intact, but `Context.BaseURL()` always delegated to `Request.BaseURL()`, which only constructs `scheme://host:port` with no path.

The fix is one extra check at the top of `Context.BaseURL()`:

```go
func (c *Context) BaseURL() string {
    if env.IsSingleHostMode() {
        return env.Config.BaseURL
    }
    return c.Request.BaseURL()
}
```

This makes `Context.BaseURL()` consistent with `web.BaseURL()`. Multi-tenant behaviour is unchanged.

A test covering the single-host-with-path case is included.

Fixes #1452.

you know the codebase far better than I do, happy to adjust anything here.


mattwoberts:
@tejasae-afk Bit slow to responsd to this, sorry. 

Seems reasonable - setting a path in baseURL isn't probably very common hence it slipped through. I can't think of any issues / reasons that this is a bad idea...

Issue #1500, opened 2026-04-15
[BUG] blockquote element is missing


JimKnoxx:
**Fider Cloud or Self Hosted**
master version for internal testing

**Describe the bug**
Before the current master a quote (with the > character) was a blockquote html element, no matter if markdown editor or richtext.
In the current master this will just render as a '>' character inside a p html element, no matter if markdown editor or richtext.
We noticed this, because we added custom styling for quotes.

**To Reproduce**
I used the demo fider server for the older version and the current master for the current version

**Expected behavior**
Quotes are a blockquote element.

JimKnoxx:
Would it be feasible to only escape `<` in this case?  
In #1495 you disabled both, that breaks the quoting.

mattwoberts:
Done!

PR #1504, opened 2026-04-16
Fixing quote functionality, adding quote icon to the comment editor

JimKnoxx:
**Issue:** #1500 

In #1495 a XSS vulnerability was fixed by sanitizing the editor output. This broke quoting.

I propose to only quote the `<` character in this case, so quoting still remain in tact. I am unsure though, if this would open possible xss vulnerabilities again. 

Also adding a dedicated quoting button to the editor for better UX.


mattwoberts:
Thanks @JimKnoxx - #1495 had been pulled into main and I was going to test all this stuff with it before releasing to cloud and updating the verion - you beat me to it! I did suspect this might happen... 

I'll take a look at the issue, and your proposed fix too

mattwoberts:
@JimKnoxx I think this is OK - the original security issue still seems secure with this change you made, so I think we're OK to bring this in

Issue #1501, opened 2026-04-15
Show the current Fider version


JimKnoxx:
I am not sure, if this is by design, but atm there is no way of telling the used fider version, besides logging into the docker container.

If this is something that you would be fine to change this might be a place for it?

<img width="296" height="137" alt="Image" src="https://github.com/user-attachments/assets/fedbc637-4f32-4152-ae4a-6842c950673f" />

mattwoberts:
Hi @JimKnoxx ... I'm fine with the principle of putting the version somewhere... but we'd rather is was a less prominent place, perhaps somewhere in admin settings?

JimKnoxx:
Sure! Will look into that tomorrow. Let me know if you have a specific place you would be okay with!

JimKnoxx:
@mattwoberts I did not really find a good place in the the Admin settings. 
Maybe a dedicated "about" page in the user dropdown could be a idea that opens kind of a "wanna contribute, homepage, github" information side?
Or just show the number by the powered by only for Admins / Contributers, not for "normal" users and visitors?

mattwoberts:
Or - this is obviously something that's quite useful for the self-hosting open source community, but not something we want on the cloud version, so you could wrap it in a check for if we're the cloud version (IsMultiHostMode will prb suffice)

And could you make the version not part of the link - some text right next to it?

JimKnoxx:
> you could wrap it in a check for if we're the cloud version

Will do!

> And could you make the version not part of the link - some text right next to it?

@mattwoberts sure, but I guess you want to keep the bold as part of the link? So it would look like this:

<img width="191" height="59" alt="Image" src="https://github.com/user-attachments/assets/18d74cec-4ef4-47e3-a9b4-2dd5160abe2d" />

Maybe place it below? 

<img width="149" height="65" alt="Image" src="https://github.com/user-attachments/assets/a37a15ae-43a3-468a-9b85-1e4751ea4572" />

mattwoberts:
Below is good!

PR #1505, opened 2026-04-16
Add Fider version to Powered by Fider

JimKnoxx:
**Issue:** #1501 

If this is not a feature you would want to have in fider, feel free to close this PR.
This will show the version (hopefully, only have 'dev' as version) next to the `Powered by Fider` string.
If the instance is running at dev, it will not show a version.

### DIFF
--- a/public/components/common/PoweredByFider.scss
+++ b/public/components/common/PoweredByFider.scss
@@ -5,13 +5,8 @@
 
   a {
     color: var(--colors-blue-700);
-    font-size: 12px;
   }
   a:hover {
     color: var(--colors-gray-900);
-  }
-  span {
-    display: block;
-    font-size: 12px;
   }
 }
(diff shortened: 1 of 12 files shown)


JimKnoxx:
@mattwoberts This is ready for review now :)

public/components/common/PoweredByFider.scss
```diff
@@ -10,4 +10,8 @@
   a:hover {
     color: var(--colors-gray-900);
   }
+  span {
+    display: block;
+    font-size: 12px;
+  }
```
mattwoberts: You probably could have used the utility classes but I'll let you off ;)

mattwoberts (approved):

public/components/common/PoweredByFider.scss
JimKnoxx: Good point!

mattwoberts (approved):

PR #1507, opened 2026-04-18
Fixed users tab in administrative view not being highlighted when active

JimKnoxx:
Found a little bug: The users tab in the administrative view was no longer being visible marked as the "active" tab.
Screenshot 2026-04-18 at 07 31 59

A different solution could also be, that you rename the menu item to members, because the heading of this tab talkes about members. But I didn't want to change the user interface here to keep the solution simple and straight forward.


mattwoberts:
Thanks @JimKnoxx , nice find :)

Issue #13, opened 2017-01-28
GET TenantByDomain should return list of boards


goenning:
Current `GetTenantByDomain` operation should also return the list of available boards for given tenants. A new table is needed for this. Board should contain it's Id, Name, IsDefault.

goenning:
As of v0.1, we'll not have boards.

PR #1513, opened 2026-04-28
fix(email): unescape HTML entities in rendered subject

americodias:
## Bug

Email subjects rendered with characters like `+`, `"`, `&`, `<`, `>` arrive in clients as their HTML numeric entities. Real-world example seen on a self-hosted instance:

> Subject: `[Plataforma da Encontrar&#43;se] Two columns of &#34;status&#34;`

Any site name or post title containing one of those characters is affected — quotes in titles are common, so most non-trivial use is hit.

## Cause

The subject template (`{{define "subject"}}…{{end}}`) shares the template set with the HTML body and is rendered via `html/template`. Go's `htmlReplacementTable` escapes `+ " ' & < >` to numeric entities (UTF-7 / JS-injection defense). Correct for HTML body, wrong for the SMTP `Subject:` header — that header is plain text and clients do not HTML-decode it.

## Fix

Run `html.UnescapeString` on the extracted subject inside `RenderMessage`. Body rendering is unchanged.

## Test

`TestRenderMessage_SubjectUnescapesHTMLEntities` posts a name containing `+ " & < >` through the existing `echo_test` template and asserts the literal characters survive in the subject. Verified to fail on `main` and pass with the fix.

### DIFF
--- a/app/services/email/email_test.go
+++ b/app/services/email/email_test.go
@@ -10,6 +10,20 @@ import (
 	. "github.com/getfider/fider/app/pkg/assert"
 )
 
+func TestRenderMessage_SubjectUnescapesHTMLEntities(t *testing.T) {
+	RegisterT(t)
+
+	// Reproducer for the issue where characters like '+', '"', '&', '<', '>'
+	// in the rendered subject ended up as their numeric HTML entities
+	// (e.g. "&#43;", "&#34;") because the subject is rendered through
+	// html/template along with the body. SMTP headers are plain text and
+	// must contain the literal characters.
+	message := email.RenderMessage(context.Background(), "echo_test", email.NoReply, dto.Props{
+		"name": `Encontrar+se "quoted" & <tagged>`,
+	})
+	Expect(message.Subject).Equals(`Message to: Encontrar+se "quoted" & <tagged>`)
+}
+
 func TestRenderMessage(t *testing.T) {
 	RegisterT(t)
 
--- a/app/services/email/message.go
+++ b/app/services/email/message.go
@@ -3,6 +3,7 @@ package email
 import (
 	"bytes"
 	"context"
+	"html"
 	"mime"
 	"strings"
 	"unicode"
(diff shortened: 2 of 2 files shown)


mattwoberts:
Thanks for this. I knew that tweaking a few things for security would cascade down to some issues like this - it's a can of worms. I'm pretty sure your fix doesn't introduce any security issues....

americodias:
Confirmed safe — went through it explicitly:

**Scope:** only the value emitted into the SMTP `Subject:` header. Body rendering (where `html/template` is the actual XSS defense) is untouched.

1. **HTML/JS in subject.** `html.UnescapeString` brings back `< > & " ' +`. SMTP `Subject:` is plain text — no HTML/JS rendering. The subject isn't reflected back into HTML elsewhere in the codebase that I could find, so no XSS path opens.

2. **CRLF / SMTP header injection.** The one I checked most carefully:
    - `html.UnescapeString` is single-pass. A malicious `Foo&#13;&#10;Bcc: …` first gets `&` → `&amp;` by `html/template`, ending up as `Foo&amp;#13;&amp;#10;Bcc: …`. `html.UnescapeString` then undoes only one layer → literal text `Foo&#13;&#10;Bcc: …`, no actual CR/LF.
    - I verified the single-pass behaviour empirically — even a doubly-encoded input `Foo&amp;#13;&amp;#10;Bcc: x` (escaped to `Foo&amp;amp;#13;&amp;amp;#10;Bcc: x`) decodes back to literal `Foo&amp;#13;&amp;#10;Bcc: x` — still no CR/LF.
    - **Raw** `\r` / `\n` already pass through `html/template` unchanged (it doesn't escape control chars), so they reach the subject in *both* pre- and post-fix code. If raw CR/LF in subjects is a concern, it pre-exists this PR and isn't introduced or widened by the unescape.

3. **UTF-7 vector.** The original `+` escape was a defense for HTML parsing contexts where charset switches could turn `+` into a JS-injection sequence. `Subject:` is plain text, no HTML parsing; not relevant here.

Adjacent thought worth a follow-up PR (separately): defense-in-depth would be to reject raw CR/LF in `EncodeSubject`. Happy to do that as a second PR if useful — just say the word.

PR #1514, opened 2026-04-29
Make post search respect active filters

lol2x:
## Summary

On the Home page, applying a status (or tag/my-votes/my-posts/untagged) filter and then typing in the search box previously dropped every filter — search returned posts of all five statuses with no tag/ownership filtering. The filter UI also disappeared the moment the search input got focus, so users had no way to see or change what was active.

This PR aligns the search branch of `searchPosts()` with the no-query branch so filters are honored consistently, and keeps the filter sidebar visible during search. Score-based ordering during search is preserved (the sort selector is hidden because it has no effect when results are ranked by `ts_rank_cd`).

- `getViewData()` now takes a `tagsPlaceholder int` so both call sites can use it (filter mode binds tags as `\$3`, search mode as `\$4`).
- The search branch calls `getViewData(*q, 4)` to get the same `condition` and `statuses` the no-query branch uses, applies `MyPostsOnly` the same way, wraps the OR full-text predicate in parens to avoid OR/AND precedence dropping filter rows, and binds `pq.Array(q.Tags)` as `\$4` only when present.
- `<PostFilter>` renders unconditionally; only `<PostsSort>` is gated on `!query`.

Default behavior change: a search request with no explicit status filter now returns only `[Open, Started, Planned]` (matches filter-mode default) instead of all five statuses. To include `Completed`/`Declined` posts in search results, the user can set those statuses explicitly.

## Bugs found while working on this

Code review surfaced two pre-existing edge cases in the same code paths:

1. \`MyPostsOnly\` panicked on anonymous API calls (nil deref on \`user.ID\`).
2. Setting \`NoTagsOnly\` and \`Tags\` together produces silently empty results — the UI doesn't enforce mutual exclusivity, so a user can reach the contradictory state with two clicks.

Both predate this change but became more reachable now that the search branch applies the same filter logic. Addressed in #1515 (stacked on top of this PR) to keep this one focused on the original bug.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/app/services/sqlstore/postgres/common.go
+++ b/app/services/sqlstore/postgres/common.go
@@ -2,6 +2,7 @@ package postgres
 
 import (
 	"context"
+	"fmt"
 	"net/url"
 	"regexp"
 	"strings"
@@ -35,7 +36,7 @@ func MapLocaleToTSConfig(locale string) string {
 	return enum.MapLocaleToTSConfig(locale)
 }
 
-func getViewData(query query.SearchPosts) (string, []enum.PostStatus, string) {
+func getViewData(query query.SearchPosts, tagsPlaceholder int) (string, []enum.PostStatus, string) {
 	var (
 		condition string
 		sort      string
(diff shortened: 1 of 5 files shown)


mattwoberts:
@lol2x Yep, this makes sense, I like this.

mattwoberts (approved):

mattwoberts:
Superseded by #1515 — same changes plus edge-case hardening.

PR #1525, opened 2026-05-06
Update version string formatting to remove suffix

JimKnoxx:
Sorry for this fail... Haven't looked into prod data enough...

It renderd `vv0.35.0-aae615eebbc3ef6f6f81ad72bfb7978c15c84256`
v comes from the CLI and the commit hash is added too. 
so, lets ignore both, so that we can have v0.35.1 standing there (with the next release 😄)


mattwoberts:
Eek my bad too - didn't spot this either. Anyone else finding it harder to keep up with everything with all the claude stuff, it's much easier for things to fall through the cracks!

PR #1526, opened 2026-05-06
feat(home): add Recently Responded sort and i18n the Pending filter

lol2x:
## Summary
- New sort option **Recently Responded** in the home page sort dropdown — orders posts by `response_date` (when an admin last changed status), newest first.
- Filters out posts with `response_date IS NULL` so the sort stays meaningful regardless of the active status filter. Without this, picking the new sort while the default Open/Started/Planned filter is active produces a long tail of unresponded posts in arbitrary order.
- i18n the Pending status filter (collaborator/admin only) and the inline "pending" badge in `ListPosts` — both were previously hardcoded English.
- Adds Polish translations for the two new keys plus the previously-untranslated Pending key.

## Why
Admins/collaborators often want to review what's been recently dealt with — completions, declines, planned items — without having to also remember to flip the sort manually. Sorting by `id`/creation date doesn't surface this; the existing per-status `planned`/`completed`/`declined` views are deprecated in favor of status filters but those don't carry the response-date sort. This stitches that gap.

## Behavior
- Pairs naturally with a Completed + Declined status filter to get a chronological "what got done" feed.
- With default filters, the IS NOT NULL guard means only posts that have ever been responded to appear (e.g. a post reopened after being marked complete still shows because its `response_date` is preserved).
- Filter and sort remain independent — picking this sort doesn't override your status selection.


mattwoberts:
Sorry, closing this too - sounds like you have the need for some tools to help collobators deal with posts. I think this is well worth a discussion on our feedback board, and we may well agree on a solution that can then be pull requested in.

PR #1527, opened 2026-05-06
feat(search): allow collaborators to find posts by response text

lol2x:
## Summary

Extend the post search so collaborators and administrators can match against
`posts.response` text (the staff note set when changing a post's status).

This makes it possible to find posts by what was written when closing them —
e.g. retrieving every post marked "Shipped in v2.3.1". For visitors and
anonymous users the SQL and behaviour are unchanged.

## How it works

- New generated `tsvector` column `posts.search_response` (response only,
  GIN-indexed). NULL when the post has no response, so the index stays small.
- `searchPosts` extends the predicate and ranking with
  `q.search_response @@ ...` only when `user.IsCollaborator()` is true.
- `findSimilarPosts` is unchanged.

## Notes

Response search uses `websearch_to_tsquery` with the raw query so
version-style tokens like `v2.3.1` are preserved as a single token. The
existing title/description path still uses the prefix-matching `to_tsquery`
flow.

No frontend changes — the search box automatically benefits from the wider
scope when used by a logged-in collaborator/administrator.

### DIFF
--- a/app/services/sqlstore/dbEntities/post.go
+++ b/app/services/sqlstore/dbEntities/post.go
@@ -18,6 +18,7 @@ type Post struct {
 	Description    string         `db:"description"`
 	CreatedAt      time.Time      `db:"created_at"`
 	Search         []byte         `db:"search"`
+	SearchResponse []byte         `db:"search_response"`
 	User           *User          `db:"user"`
 	HasVoted       bool           `db:"has_voted"`
 	VotesCount     int            `db:"votes_count"`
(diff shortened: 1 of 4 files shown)


mattwoberts:
I don't think this is a bad implementation, but I'm going to close it because it's not something on our roadmap, or something we've discussed and agreed is a good addition for everyone. I'd recommend you post this as a suggestion on [our feedback board](https://feedback.fider.io) and see if we get much traction.

PR #1542, opened 2026-05-18
Roadmap view

mattwoberts:
New Roadmap view, using the posts status to determine which roadmap column to show the posts in


JimKnoxx:
Nice feature! Thank you for working on this.
Had a quick look into this. Little feedback, hope you don't mind while in dev:
I would flip the moderation header information and the "user toggles / menus" in the header. Currently, once there is something to moderate it will place the user toggles / menus on the next line (on a small screen). Flipped it would place the moderation information on the next line.

Or I think the better user flow would be to place the moderation info over/under the "your feedback..." button, so it is grouped with the thing that triggers the moderation.

mattwoberts:
Hey @JimKnoxx 

Thanks for the feedback - although I'm not sure I agree ...

Here is a small screen on the current fider:

CleanShot 2026-05-19 at 20 50 57@2x


And with the new roadmap:

image


The content moderation button remains underneath, which I think works. And the new nav links remain as part of the navigation, which I would say makes sense since they're part of the nav header and not something you'd want to see move down on smaller screens?

JimKnoxx:
Hah well, then my browser had the exact width for it to happen:
image

I have a 16:9 monitor but vertical and also using vertical taps (Zen Browser).

mattwoberts:
Cheers @JimKnoxx I'll take a look at that

Issue #1535, opened 2026-05-10
[BUG] no char limit on comment section leads to super long comment


Raunaksplanet:
**Fider Cloud or Self Hosted**
https://create.fider.io/posts/612/value-0d-0ax-injected-20burp-ai-test

**Describe the bug**
a super long comment with no char limit can be added

**To Reproduce**
1. navigate to self hosted fider instance 
2. create a new suggestion
3. enter super long comment
4. observe that a user can add super long comment

**Expected behavior**
there should be a char limit so a user cant spam 

**Screenshots**
Proof of concept https://create.fider.io/posts/612/value-0d-0ax-injected-20burp-ai-test#comment-52820

mattwoberts:
Fixed in #v0.36.1

PR #1543, opened 2026-05-18
fix(comments): unify create + edit limit at 2000 chars (#1535)

mattwoberts:
## Summary
- Unifies the comment character limit at **2000** for both create and edit paths. The prior commit on main set `AddNewComment` to 4000 and `EditComment` to 2000, which would silently block users from editing comments they were allowed to post.
- Updates the matching frontend constants (`DEFAULT_MAX_LENGTH` in `CommentEditor`, `COMMENT_MAX_LENGTH` in `CommentInput`, the explicit `2000` passed in `ShowComment`) so the live counter, disabled-submit gating, and backend rejection all agree.
- Adjusts the comment validation tests to the new boundary.

Addresses GitHub issue #1535 (no character limit on comments enables spam).

## Test plan
- [ ] `make lint` clean
- [ ] `godotenv -f .test.env go test ./app/actions` passes (incl. `TestAddNewComment_TooLongContent`, `TestAddNewComment_AtMaxLength`, `TestEditComment_TooLongContent`)
- [ ] `make watch`: open a post, paste 2001 chars into the comment editor — counter goes red at the limit, Post button disables
- [ ] Backspace to 2000 — Post enables, submit succeeds
- [ ] Edit an existing comment past 2000 chars — Save disables; counter red
- [ ] Switch to markdown mode — textarea hard-caps at 2000 chars natively
- [ ] Force-bypass via DevTools — backend returns 400 with i18n error on `content`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/app/actions/post.go
+++ b/app/actions/post.go
@@ -221,8 +221,8 @@ func (action *AddNewComment) Validate(ctx context.Context, user *entity.User) *v
 
 	if action.Content == "" {
 		result.AddFieldFailure("content", propertyIsRequired(ctx, "comment"))
-	} else if len(action.Content) > 4000 {
-		result.AddFieldFailure("content", propertyMaxStringLen(ctx, "comment", 4000))
+	} else if len(action.Content) > 2000 {
+		result.AddFieldFailure("content", propertyMaxStringLen(ctx, "comment", 2000))
 	}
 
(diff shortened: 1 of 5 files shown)


mattwoberts:
Closing — wrong direction. Will open a new PR that aligns edit limit up to 4000 instead.

PR #1562, opened 2026-06-02
Added OAUTH_REDIRECT_URI_USE_BASE_URL env config

anderslarssonvbg:
**Issue:** In the current code when using multi tenant mode, the oauth redirect uri is hard coded to a url which begins with https://login. followed by {tenant_cname}. If that is not wanted, the OAUTH_REDIRECT_URI_USE_BASE_URL env variable can now be set to "true" and then the redirect uri will then be just https://{tenant_cname}


mattwoberts:
Curious how you're using fider in multi mode that requires this change?

anderslarssonvbg:
I have set up Fider internally for a customer where they wanted to handle feedback for several internal systems separately, without the need of several instances of Fider. So they have subdomains like feedback.xxx.zzz.com and feedback.yyy.zzz.com. The hardcoded login. domain in the code then requires also login.feedback.xxx.zzz.com and login.feedback.yyy.zzz.com domain, with own SSL certificates for these, which is an additional cost and administration.

mattwoberts:
> I have set up Fider internally for a customer where they wanted to handle feedback for several internal systems separately, without the need of several instances of Fider. So they have subdomains like feedback.xxx.zzz.com and feedback.yyy.zzz.com. The hardcoded login. domain in the code then requires also login.feedback.xxx.zzz.com and login.feedback.yyy.zzz.com domain, with own SSL certificates for these, which is an additional cost and administration.

Wouldn't you be better off hosting on one domain - zzz.com for example, so you just need the one login.zzz.com, then you have feedback1.zzz.com and feedback2.zzz.com?

anderslarssonvbg:
Unfortunately due to internal restrictions, I'm just allowed to use single level domains under yyy.zzz.com and xxx.zzz.com, and login is already used there for some of them. So what I have done and what the MR contains is the possibility to skip the login domain, it works perfectly without it.

But this can probably be solved by introducing a separate variable where the oauth redirect url can be specified, this solution was just coded to not interfere with the current behaviour.

mattwoberts:
I'm torn on this - it's valid, and gets you out of a sticky situation, but it adds additional config that I don't imagine will be of any use to anyone else

Hmmmm

anderslarssonvbg:
But it's an optional config parameter, no one needs to set it if they don't want to?

Another solution could be a config parameter where you can specify what domain the redirect should go to, instead of the hard coded "login.", defaulted to "login." but which I could set to "" ?

PR #1564, opened 2026-06-08
feat(tenant): add description template field to tenant settings

jeffreyvdhondel:
## What & why
  
  Adds an optional, per-tenant **Idea Template** that prefills the description editor when a user opens the "Share your idea" modal. Admins configure a structured template (e.g. _What is your idea? / What problem does 
  it solve? / Who benefits?_) so submissions arrive with consistent, useful detail instead of a blank box.

  The setting follows the same pattern as the existing `WelcomeMessage` / `Invitation` tenant fields, so it slots into the current architecture with no new concepts.

  ## Behaviour

  - **Admin → General Settings** gains an "Idea Template" textarea (Markdown supported).
  - When a template is configured, the new-post description editor opens **prefilled** with it, and title auto-population is suppressed so the template heading doesn't leak into the title.
  - **No template configured → empty editor** (unchanged default experience).
  - An in-progress cached draft always takes precedence over the template, so drafts are never overwritten.

  ## Changes

  **Backend** :

  - Migration `202606081200_add_tenant_description_template.sql` adds `tenants.description_template TEXT NOT NULL DEFAULT ''`
  - `description_template` threaded through the tenant entity, DB scan struct + `ToModel`, the two tenant `SELECT`s, the settings `UPDATE`, the `UpdateTenantSettings` command/action (incl. a 2000-char validation), and
  the admin handler

  **Frontend:**

  - `descriptionTemplate` added to the `Tenant` model and the update-settings request payload
  -  New "Idea Template" field on the General Settings admin page
  - `ShareFeedback.tsx` prefills the editor from the tenant template when no draft exists

  ## Notes

  - Backwards compatible: column defaults to `''`, so existing tenants behave exactly as before until they opt in.


mattwoberts:
Sounds like a good idea. Is there an issue or fider suggestion for this ?

mattwoberts:
I think this? https://feedback.fider.io/posts/474/allow-custom-suggestion-templates

jeffreyvdhondel:
Nice let me fix the merge conflict!

PR #1579, opened 2026-06-22
feat(webhooks): ALLOW_PRIVATE_NETWORK_TARGETS env to allow internal targets

hodyhq:
Self-hosted Fider instances frequently need webhook targets on the same LAN (Plane, n8n, internal Slack bridges, etc.). Upstream's SSRF guard in WebhookURL() rejects every RFC1918 / loopback / link-local target with no override, which makes Fider unusable for internal-only deployments that depend on webhook integrations.

This commit adds an opt-in env var WEBHOOK_ALLOW_PRIVATE_IPS (default false) that, when set, skips the private-network check. Format validation (well-formed URL, http/https scheme) still runs.

Tests cover: the new env keeps existing blocked addresses blocked when unset (no behavior change for hosted Fider); when enabled, RFC1918, loopback, link-local, and "localhost" all pass validation; opting in still rejects non-http schemes and malformed URLs.

**Issue:**


mattwoberts:
Should be an issue really, but I'll let you off

mattwoberts:
Thanks @hodyhq , I'll bring this in now 👍

PR #1580, opened 2026-06-22
feat(post-status): add Review status between Open and Planned

hodyhq:
Adds a new built-in post status "Review" for the triage step where an Open post is being evaluated before being accepted onto the roadmap (Planned). Same blue lozenge family as the other in-lifecycle statuses (Open/Planned/Started); distinct icon (eye) so the bubble reads unambiguously.

Changes:
- enum.PostReview = 7 added to the Go enum, registered in both name<->ID maps.
- SetResponse validation tightened to allowlist-by-name: any unknown status string and PostDeleted (reserved for deletion) are rejected; Review now passes. Replaces the prior numeric range check.
- Postgres default-view filter and "all" view filter include Review so triaged posts show alongside Open/Started/Planned.
- TypeScript PostStatus mirror gets the new Review constant and is inserted into PostStatus.All in workflow order (Open -> Review -> Planned -> Started -> Completed -> Duplicate -> Declined).
- ShowPostResponse lozenge: HeroIconEye + same blue swatch as the other open-lifecycle statuses; Trans key for the translated label.
- locale/en client + server: "Review" translation.
- New test TestSetResponse_ReviewIsValid; the existing TestSetResponse_InvalidStatus (PostDeleted) still passes via the allowlist check.

Use case: organizations that want a triage step ("is this worth making a project?") before promoting an Open idea onto a roadmap. The new status is fully optional - existing posts default to Open and no migration of in-flight posts is required.

**Issue:**

### DIFF
--- a/app/actions/post.go
+++ b/app/actions/post.go
@@ -257,7 +257,9 @@ func (action *SetResponse) IsAuthorized(ctx context.Context, user *entity.User) 
 func (action *SetResponse) Validate(ctx context.Context, user *entity.User) *validate.Result {
 	result := validate.Success()
 
-	if action.Status < enum.PostOpen || action.Status > enum.PostDuplicate {
+	// Only known statuses are settable via this action. PostDeleted is reserved
+	// for the deletion flow, never set through SetResponse.
+	if action.Status == enum.PostDeleted || action.Status.Name() == "unknown" {
 		result.AddFieldFailure("status", propertyIsInvalid(ctx, "status"))
 	}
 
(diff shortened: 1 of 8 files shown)


mattwoberts:
I actually don't think this is a bad idea, but there's no accompanying issue and since this is a feature / functionality change, it needs to have had prior discussion on [our feedback site](https://feedback.fider.io) 

For this reason I'm going to close for now, sorry.

hodyhq:
Thanks for the feedback. Was actually already working to change this to the Ability to create Custom Statuses, I'm submitting that PR today. it actually references "https://feedback.fider.io/posts/111/allow-admins-to-customise-list-of-statuses"

PR #1582, opened 2026-06-22
feat(ResponseLozenge): show response detail inline with status pill

hodyhq:
When an admin/collaborator responds to a post with status text (e.g. "In beta testing", "Awaiting customer feedback", or a label from an external integration), the status lozenge stays terse ("Started") and the response body only appears further down the page. This makes it hard to scan a list of posts and see *why* each one is in its current state.

This change extracts the first non-empty line of `response.text` (stripping simple markdown bold/italic markers and truncating at 60 characters) and renders it inline inside the same status pill, after a middle-dot separator:

  Started · In beta testing

The substage is wrapped in a `.c-status-substage` span (80% opacity, preserved casing) so the primary status remains the visual headline. Renders on small / xsmall / default lozenge sizes only — the micro chip stays clean.

Zero-config: works for any response text, whether typed by an admin or written by an external integration that pushes status updates via the API. The full response body still renders below in ResponseDetails as before, so no information is hidden.

**Issue:**

### DIFF
--- a/public/components/ShowPostResponse.tsx
+++ b/public/components/ShowPostResponse.tsx
@@ -100,10 +100,23 @@ const getStatusTranslation = (status: PostStatus): JSX.Element => {
   }
 }
 
+// Extracts the first non-empty line of the response markdown — that's where
+// the Plane webhook writes the substage label (e.g. "In Beta Testing").
+const extractSubstage = (text?: string): string | null => {
+  if (!text) return null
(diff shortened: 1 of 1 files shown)


mattwoberts:
Sorry, closing. Because there isn't a suggestion with traction on our feedback site (https://feedback.fider.io) and it's a change to the UI, it's not something we'd accept without prior approval. Would you mind creating something there and opening a discussion to see if this is something we'd want to add?

